PR #3, opened 2015-09-30
HV Control Updates

BenLand100:
This is awaiting additions to the XL3 status packet to get the init status of the XL3 so that HV controller can enter into a known state as soon as the XL3 connects.


BenLand100:
Obviously will have to TIAS but this should be good to go.

tlatorre-uchicago:
TIAS?

BenLand100:
Try It And See - forgot to expand my acronyms.

tlatorre-uchicago:
Ah OK! I thought it was "test it at site".

PR #4, opened 2015-10-08
Adding TUBii into ORCA for the first time.

icoulter:
This commit is NOT meant to be merged. Its purely for discussion purposes.

This is the first addition of a TUBii object. Basically it has some textboxes and buttons which can be used to control the pulser and delay functions of TUBii. Rates, number of pulses and delay lengths can be set and sent as commands via ethernet to TUBii.
This has been tested here at Penn and seems to work but I would be surprised if was doing things in the most efficient way.


Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiModel.m
```diff
@@ -0,0 +1,302 @@
...
+    NSString* command=@"syncpulser 0 0 \r\n";
+    [self connectToPort:command];
+}
+
+- (void) setSmellieDelay:(float)value
+{
+    smellieDelay = value;
+}
+
+- (void) loadSmellieDelay:(BOOL)button
```
cjones2500: @icoulter I thought the SMELLIE delay was originating from the MTC/D? I have no objections to the Tubii being able to do this but I don't think it should be an ability you can access through the TUBii GUI (even though it originates from TUBii)

Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiModel.m
```diff
@@ -0,0 +1,302 @@
...
+}
+
+// Initialize the model.
+// Note that this is initWithCoder and not just init, and we
+// call the superclass initWithCoder too!
+- (id) initWithCoder:(NSCoder *)aCoder {
+    self = [super initWithCoder:aCoder];
+    
+    // Initialize model member variables
+    smellieRate = 0;
```
cjones2500: @icoulter You can access these from the SMELLIE/TELLIE Model and maybe we should make these subsystems provide a default value ? Also I'm not sure we should give control of SMELLIE to the TUBii Model ? What are your thoughts?

cjones2500:
@icoulter hey it seems reasonable what you have at the moment. There are some questions about which model should have control of what concerning TELLIE and SMELLIE? I think SMELLIE/TELLIE should request from the TUBii model to ask for certain trigger rates/number of triggers (which you have) but TUBii should not be able to send SMELLIE/TELLIE triggers (in its GUI). I can't think of a use case when you would risk this will SMELLIE. It may be helpful for debugging with TELLIE but for SMELLIE this functionality should be locked out unless SMELLIE explictly asks for a trigger.

Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiModel.m
icoulter: Chris, this is an example to check we can talk to TUBii and weren't making any major mistakes in the code. This is not what will be added to the master copy of ORCA.
It was decided on the calls that we should have a TUBii model that first contains everything TUBii can do. After that we can worry about the integration of these into ELLIE, MTC, etc...

Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiModel.m
icoulter: No, the SMELLIE delay with originate from TUBii. As will the delay used by TELLIE.
The argument of whether these are hardcoded, open to all or (most likely) open to certain expert/calibration users is a debate for later as, like I say, this is just a prototype.

icoulter:
We can debate the exact use later but TUBii was asked to provide a delay and a pulser for both SMELLIE and TELLIE. The final control can be added after the functionality is in place.

tlatorre-uchicago:
I think this looks good so far. I am working on creating a class to interface with all of the servers that support the redis protocol. So far my idea is that the object will be created inside the `init` method like

```
- (void) init
{
    self.server = [[RedisServer alloc] initWithHost:"tubii" port:4001 timeout:2];
}
```

and then you can send commands that should just return "+OK" by

```
[self.server okCommand:"tellie_delay %f", 1.0];
```

which will raise an exception if the command fails or you get a non "+OK" response. For integer responses:

```
uint32_t delay = [self.server intCommand:"get_delay"];
```

icoulter:
Great. That seems pretty reasonable. I will work with that in mind.

tlatorre-uchicago:
@icoulter is this superseded by pull request #21? If so, can I close it?

icoulter:
Yes. #21 incorporates much of this and supersedes the rest.

PR #6, opened 2015-11-10
Fixed memory leaks pointed out by Mark Howe

EdLeming:
You'll notice a couple of edited lines where I've commented out autorelease pools where I don't think they're necessary, if possible it would be good to get some feedback on if this is correct. As far as I understand they're only best practice if you're creating a load of objects in a loop, or when you spawn a thread.

### DIFF
--- a/Source/Objects/Custom Hardware/SNO+/ELLIE/ELLIEModel.m
+++ b/Source/Objects/Custom Hardware/SNO+/ELLIE/ELLIEModel.m
@@ -116,7 +116,7 @@ smellieDBReadInProgress = _smellieDBReadInProgress;
     /* 
      Start run using run control object and push initial TELLIE run doc to telliedb.
      
-    Possible additions:
+     Possible additions:
         Use SNOPModel to check if tellie run type is masked in
      */
     
@@ -135,7 +135,7 @@ smellieDBReadInProgress = _smellieDBReadInProgress;
 {
     /*
      Use run control object to stop a tellie run.
-     */
+    */
     
     //add run control object
     NSArray*  runControlObjsArray = [[(ORAppDelegate*)[NSApp delegate] document] collectObjectsOfClass:NSClassFromString(@"ORRunModel")];
@@ -149,8 +149,8 @@ smellieDBReadInProgress = _smellieDBReadInProgress;
 -(void) pollTellieFibre
 {
     /*
-    Poll the TELLIE hardware using an XMLRPC server and requests the response from the
-    hardware.
+     Poll the TELLIE hardware using an XMLRPC server and requests the response from the
+     hardware.
     */
     
     NSString* responseFromTellie =[self callPythonScript:@"/Users/snotdaq/Desktop/orca-python/tellie/tellie_readout_script.py" withCmdLineArgs:nil];
@@ -238,7 +238,7 @@ smellieDBReadInProgress = _smellieDBReadInProgress;
 -(void) stopTellieFibre:(NSArray*)fireCommands
 {
     /*
-        Call tellie stop script. The script itself is stored on the DAQ1 machine.
+     Call tellie stop script. The script itself is stored on the DAQ1 machine.
     */
     NSString *responseFromTellie =[self callPythonScript:@"/Users/snotdaq/Desktop/orca-python/tellie/tellie_stop_script.py" withCmdLineArgs:nil];
     NSLog(@"Response from Tellie: %@\n",responseFromTellie);
@@ -755,7 +755,7 @@ smellieDBReadInProgress = _smellieDBReadInProgress;
     [runControl performSelector:@selector(haltRun)withObject:nil afterDelay:.1];
 }
 
--(NSNumber*) fetchRecentVersion
++(NSNumber*) fetchRecentVersion
 {
     /*
      Query smellie config documenets on the smelliedb to find the most recent config versioning
@@ -791,7 +791,7 @@ smellieDBReadInProgress = _smellieDBReadInProgress;
     return currentVersionNumber;
 }
 
--(NSMutableDictionary*) fetchCurrentConfigurationForVersion:(NSNumber*)currentVersion
++(NSMutableDictionary*) fetchCurrentConfigurationForVersion:(NSNumber*)currentVersion
 {
     /*
      Fetch the current configuration document of a given version number.
--- a/Source/Objects/Custom Hardware/SNO+/ELLIE/ELLIEModel.m
+++ b/Source/Objects/Custom Hardware/SNO+/ELLIE/ELLIEModel.m
@@ -4,6 +4,9 @@
 //
 //  Created by Chris Jones on 01/04/2014.
 //
+//  Revision history:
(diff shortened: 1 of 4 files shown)


Source/Objects/Custom Hardware/SNO+/ELLIE/ELLIEModel.m
```diff
@@ -610,7 +607,7 @@ - (void) couchDBResult:(id)aResult tag:(NSString*)aTag op:(id)anOp
 
 -(void)startSmellieRunInBackground:(NSDictionary*)smellieSettings
 {
-    NSAutoreleasePool *pool = [[NSAutoreleasePool alloc] init];
+    NSAutoreleasePool *pool = [[NSAutoreleasePool alloc] init]; //Not sure what this is doing? Ed
```
cjones2500: So I may be incorrect here but I thought this would wrap all of the objects within the selector performed on the main thread (see next line) into an Auto release pool and thus I wouldn't have to worry about these objects. This may be incorrect, I didn't get around to checking it.

cjones2500:
@EdLeming please make sure that you check all the functionality with the TELLIE and SMELLIE sub system computers turned on when you upgrade this to the master/production branch of Orca. We should not rely upon subgroups software behaving properly (out of principle) and make sure Orca doesn't send any rubbish to subsystems. I know SMELLIE/TELLIE have protections but we should make sure Orca behaves as expected. I'm not sure about the "drain" replacements. "Drain" is an old version of "release" that is only applicable for NSAutoRelease objects. "Release" will reduce the reference count on any object whereas "drain" will only do this on NSAutoRelease objects (see http://stackoverflow.com/questions/1659545/difference-between-drain-release-dealloc-and-retain-in-objective-c )

Source/Objects/Custom Hardware/SNO+/ELLIE/ELLIEModel.m
BenLand100: My understanding is that only objects explicitly added to the autorelease pool by calling the '''autorelease''' selector on the object would be released when the pool is released explicitly or drained explicitly. Creating a new autorelease pool simply means all objects that have '''autorelease''' invoked after that pool was created go into that pool. So this pool is doing something but not automagically handling all allocated memory.

Source/Objects/Custom Hardware/SNO+/ELLIE/ELLIEModel.m
```diff
@@ -465,7 +468,7 @@ -(void) _pushInitialTellieRunDocument
         [NSThread sleepForTimeInterval:0.1];
     }
     
-    [runDocPool release];
+    [runDocPool drain];
```
BenLand100: This one should probably be a ```release``` since the pool was created in this method.

BenLand100:
This is purely speculative but I wonder whether we would be better off avoiding ```NSAutoreleasePool``` altogether. Objects in objective-c are reference counted, and when the count reaches zero the memory is freed. Under ORCA's configuration the reference counting is done manually with ```retain``` and ```release``` (as opposed to the ARC system that is used in more modern code bases which does all this automatically...). The ```NSAutoreleasePool``` is a mechanism for being lazy about calling ```release``` where instead one adds objects to the 'inner-most' pool by invoking ```autorelease``` on the objects. The pool then invokes ```release``` on each object in the pool when either ```drain``` (multiple times) or ```release``` (only once - also frees the pool) is invoked on the pool. There is no explicit reason one needs the pools as one can always call ```retain``` and ```release``` when appropriate. 

I am only concerned because I see cases where ```autorelease``` is being invoked on e.g. function results, and it is entirely unclear which pool with what scope they are being added to and what the lifetime of the memory associated with these objects actually is. This could lead to segfaults if memory is freed too early or memory leaks if the pools are never drained or released.

tlatorre-uchicago:
@BenLand100 agreed. I'm not sure why these aren't just being freed after use. Is it because the couch document that is being added happens asynchronously in another thread and so needs to live beyond the end of the function call?

If so, I see two solutions. One, I think the save operation should just block. There is too much code in our models right now that is doing things asynchronously that should be done with a blocking call and a timeout. If the CouchDB database needs to be updated or a piece of hardware needs to change before a run start, then we should *block* on that call and if it times out, then depending on how important it is, we abort the run or continue.

The other option is to set up a function which is called after the document is posted which frees the dictionary, but this is more complicated than it needs to be, and if the document needs to be posted for data quality issues, then it should block with a timeout.

I use the example of posting to the couch database because that is what most of the examples of the autoreleasepool seem to be for: create a dictionary on the stack and send it to update a couch document.

Another comment: there are still a lot of places in the code with huge chunks of commented out code. If these are useful bits of code, we should just make them into a separate function and uncomment them, otherwise just delete it. It's confusing later to have to go through the code and not understand why a block is commented out.

EdLeming:
Thanks for the comments so far. 

Tony: Could you elaborate a little on what you mean by block? Do you have an example already coded into orca? It sounds like you're talking about serialisation of processing, but the only objective-C documentation I can find on 'block' is a type of function with a little added memory cleverness

https://developer.apple.com/library/ios/documentation/Cocoa/Conceptual/Blocks/Articles/00_Introduction.html#//apple_ref/doc/uid/TP40007502

tlatorre-uchicago:
No, I just mean block in the sense of the function doesn't return until the document is posted to CouchDB, in which case the dictionary doesn't need to be kept around anymore.

I guessed that the posting to CouchDB was happening asynchronously because the addDocument method https://github.com/snoplus/orca/blob/66016f352dfc50647413f778e3f8f6f45687f738/Source/Objects/Misc%20Objects/DatabaseSupport/CouchDB/ORCouchDB.m#L148 puts something on a queue.

Not sure if there is an equivalent blocking method.

EdLeming:
I've removed the NSAutoreleasePool's. As a result I've updated the associated dictionary instances to be instantiated with dictionaryWithCapacity, rather than initWithCapacity. This way I believe they are only released when all objects drop their reference to it - this may be a misunderstanding on my part? 

Re: Blocking : There doesn't seem to be a non-queuing option in the current ORCouchDB.m, I think we'd have to update it?

tlatorre-uchicago:
Hm. Yeah, I don't see one either. I can think of two options. It may be possible to sit in a while loop checking the ORCouchDBQueue size.

The other option would be to email Mark and see if he would be willing to add a blocking method. I think this is probably the best, and I think it shouldn't be too hard.

EdLeming:
After a short e-mail discussion Mark's convinced we don't need to make it this complicated. To summarise: 

Mark : 'If you pass an object off to some thread,..., you can just release the object immediately since that other object will retain and release the passed object as needed.'

In our case NSOperationQueue contained in ORCouchDBQueue is the thread that we pass off into. As a result we can release the objects in the ELLIE code immediately after passing to ORCouchDB - no blocking needed.

EdLeming:
@tlatorre-uchicago: I just checked this over. It should be good to go.

tlatorre-uchicago:
merged.

PR #7, opened 2015-12-01
Update tellie run doc fields

EdLeming:
autoreleasepools/ memory handling are tidied up in separate pull request.


tlatorre-uchicago:
Is it possible to remove the autorelease pools altogether?

EdLeming:
Yep, they are not necessary. I got rid of them all in the 'clean memory' pull request, but I'm still waiting for that one to go through. I've tidied up this specific function so you can have a look for this case.

tlatorre-uchicago:
Looks good! A couple of improvements that I think would help readability:

```
 [runDocDict setObject:docType forKey:@"type"];
 [runDocDict setObject:[NSString stringWithFormat:@"%i",0] forKey:@"version"];
```

could be written as:

```
runDocDict[@"type"] = @"tellie_run";
runDocDict[@"version"] = @"0";
```

and similarly for the other lines.

You mentioned that after the document is posted to CouchDB the dictionary should be released and the memory freed, but here you retain a reference to it in `self.tellieRunDoc`. Are you sure that it will be successfully freed eventually?

tlatorre-uchicago:
Another good thing would be a comment at the top of the function which says when this is called, what it is doing, etc. Even something simple like

```
/* Uploads the TELLIE run document to CouchDB at the start of a TELLIE run.
   If the document fails to upload, raises an Exception.
*/
```

will help a lot in the future.

EdLeming:
Cool, I've tidied up as you suggest. The self.tellieRunDoc is updated with sub run info as a run progresses and then push up at the end. It my understanding that this will get released when the object destructor gets called.

tlatorre-uchicago:
Which object's destructor? I *think* when you create a dictionary like this (without explicitly calling `alloc` `init`), the object is not guaranteed to live past the end of the life of the function unless you call `[runDocDict retain]`, at which point you need to make sure that you later call `[runDocDict release]`. If that's true then you aren't guaranteed that the dictionary will be available later in the run.

tlatorre-uchicago:
Also, it doesn't look like the function raises an Exception if the document doesn't post.

EdLeming:
Whoops, right you are,  I've now added that exception. It will raise if it takes longer than 2 seconds to post to the db. I also wrote a little tester to check if the tellieRunDoc dictionary will last beyond the end of the function call when with this method - it definitely does. As the tellieRunDoc dict is a property of the ELLIE model class it will be released when the destructor gets called on the ELLIE object. 

Apologies for the multiple commits. I added a mess of testing stuff that I was trying out this afternoon. It's all removed now.

tlatorre-uchicago:
How did you test it? I just did the same and it segfaults. In SNOPModel.m I added the following:

```
- (void) initWithCoder
{
    self->test = [NSMutableDictionary dictionaryWithCapacity:10];
    self->test[@"foo"] = @"bar";
}
```

and then later:

```
- (void) _runDocumentWorker
{
    NSLog(@"foo = %@\n", self->test[@"foo"]);
    ...
}
```

and it throws `EXC_BAD_ACCESS` at that point.

I think the solution is as simple as calling `[runDocDict release]` after you pass it off to the Orca CouchDB object.

EdLeming:
Hmm. See the attached link for test method. I simply checked that the dictionary keys were still available in the property after the function call. 

https://drive.google.com/folderview?id=0B0HOyeC4JmFNTDRJOFRseEtyQjA&usp=sharing

tlatorre-uchicago:
So, I think the issue is that when you create the dictionary with

```
NSMutableDictionary *runDocDict = [NSMutableDictionary dictionaryWithCapacity:10]
```

this is really equivalent to

```
NSMutableDictionary *runDOcDict = [[[NSMutableDictionary alloc] initWithCapacity:10] autorelease]
```

The autorelease semantics aren't extremely clear to me, but as I understand it there is an autorelease pool created on the main thread when using Apple's GUI libraries. When you create an object and set it to autorelease, it is released automatically *at the end of every event loop cycle*. See this quote from https://developer.apple.com/library/ios/documentation/Cocoa/Reference/Foundation/Classes/NSAutoreleasePool_Class/index.html:

 > The Application Kit creates an autorelease pool on the main thread at the beginning of every cycle of the event loop, and drains it at the end, thereby releasing any autoreleased objects generated while processing an event. If you use the Application Kit, you therefore typically don’t have to create your own pools. If your application creates a lot of temporary autoreleased objects within the event loop, however, it may be beneficial to create “local” autorelease pools to help to minimize the peak memory footprint. 

Therefore, the object may exist after the function call, but isn't guaranteed to live past the next event loop cycle.

There is another interesting thing with your example. You declare the property runDoc as:

```
@property (nonatomic, retain) NSMutableDictionary *runDoc
```

which means that the *generated* setter will automatically call `[self->runDoc retain]` which means that it *will* live but *only* if you use the setter, which you don't use. Note that even if you do use the setter though you will need to call `[self->runDoc release]` before you set a new runDoc.

EdLeming:
As far as I understood, by defining the property this way and using self.runDoc, I was correctly using the generated setter, which will handle the release for me when I set a new run doc. 

See here:

http://rypress.com/tutorials/objective-c/memory-management

Maybe we should have a quick call to talk about this? I appreciate all the input. It's doing a lot for my understanding.

tlatorre-uchicago:
Ah OK, so although I can't find explicit Apple documentation it seems like the retain property will not only call `retain` on an object when you use `self.runDocDict = foo` but it will *also* free any previous value, i.e. call `[self.runDocDict release]` before setting the new variable. So, using `@property (retain)` should be equivalent to

```
- (void) setRunDocDict:(NSMutableDictionary *)dict
{
    if (tellieRunDoc) [tellieRunDoc release];
    tellieRunDoc = dict;
    [tellieRunDoc retain];
}
```
in which case I think all of this will work. Wow, this stuff is really confusing. So, I think this is OK, just important to note that if you ever assign the run doc as

```
tellieRunDoc = runDocDict;
```

then it will not work.

tlatorre-uchicago:
Also, note that if you use the arrow

```
self->tellieRunDoc = runDocDict;
```

then the setter is not called! I think the arrow is interpreted as a simple attribute access like you would in C and so the setter is not called.

EdLeming:
Yeah, it's all a bit of a clusterfuck. 

OK, so I've made the creation / release of the function variables explicit to improve over-all clarity and added a comment re: the setter.

tlatorre-uchicago:
I think the comment is good. I'm now confused about explicitly calling release. I think you had me convinced that it was actually correct before, but now I'm not sure exactly what will happen when you call release. Since you already called autorelease (implicitly), I think you don't want that second release call.

EdLeming:
If you check back at the test link I sent before I've updated it for this specific case. The dict fields are still available in main. 

I believe that once the self.runDoc = runDocDict is called another reference count is added for that memory - a total of two. When I release the function variable we drop to one reference count, but we still have one so it isn't destroyed.

tlatorre-uchicago:
Where is the first reference count added? I thought that by using autorelease it defaults to having no reference count from the start, and then after `self.runDoc = runDocDict` it gets increased to 1.

EdLeming:
Yep, but I'm not autoreleasing anymore. The reference comes in the alloc.

tlatorre-uchicago:
Ah, OK. I will merge this into unstable when I get a chance.

tlatorre-uchicago:
merged. Note that I had to revert the dictionary stuff because it wouldn't compile on 5.1.

PR #22, opened 2016-01-20
HV Control Updates

BenLand100:
Somehow the previous PR disappeared... reopened for clarity.

I will probably merge and commit this to SVN later today.


BenLand100:
Additions from multistate branch conflict with unstable for three reasons:
* XIB changes (could be ignored)
 * Removed a broken outlet from Fec32
 * Removed a broken outlet in MTC model
* Project file changes (necessary to recreate)
 * Added files ORMultiStateBox.h and ORMultiStateBox.m

tlatorre-uchicago:
@BenLand100 could you send me the commit message for the svn merge?

BenLand100:
"Updates SNO+ HV control logic and gui, and adds automatic HV initialization from detector state."

tlatorre-uchicago:
Ben, I think you are going to cause the same problem with the XL3 xib that happened with the FEC32 when the PMT states boxes were added.

To fix this, I had to go through and change all instances of labelColor -> controlTextColor. See https://github.com/snoplus/orca/commit/ad98e45b637862c686f782e2987b1a0bcf2be58f

tlatorre-uchicago:
OK So I've confirmed that this is indeed the case. You can reproduce it by building with XCode 5.1.1 on OSX 10.9.5.

The labelColor -> controlTextColor thing was just a hack (I don't know what the real problem is), so maybe you can find out what the real problem is. I'll wait until you give the go ahead to try and compile this on 5.1.1 and merge.

BenLand100:
So I'm not sure I understand what could possibly be causing the issue. If you check the diff for the Fec32 xib the changes with respect to unstable were minimal - in fact the only changes were modifying the color of some text labels and changing one label from "Pending / Actual" to "Pending / Current." Does this mean the Fec32 xib in unstable also has the problem? At any rate, as minimal as the changes were I would be satisfied with taking a Fec32 xib that works and redping these changes. But then, which Fec32 xib works?

tlatorre-uchicago:
The FEC32 xib in master and unstable compiles fine. The problem is with any XML attributes that use the word "labelColor". I can think of two possible reasons that the XL3 xib has these:

- you forked before the fixes to the FEC32 xib
- when you copied it over, XCode automatically replaced controlTextColor -> labelColor

BenLand100:
I applied the hack to XL3_Link.xib - if you already have the old Xcode installed could you check that it compiles?

What I assume is happening is that newer Xcode versions use a new naming scheme for this property and for some reason that particular Xcode won't accept the new names, but I too have no concrete idea why this is a problem.

tlatorre-uchicago:
It compiles now.

tlatorre-uchicago:
@BenLand100 it looks like you've still got the old Fec32 in your branch. Could you checkout the master copy of Fec32 and then add your color changes back? I ask just because I don't want to let git merge the xib files.

BenLand100:
I pulled from master in the last few days of the DAQ workshop prior to making my changes to Fec32, and the diff against master seems to agree. Besides three lines that are presumably from being edited on my version of Xcode the only changes are related to the label changes, so it should be safe to prefer my version over master.

tlatorre-uchicago:
I'm not sure why they don't show up in the pull request diff on github, but they definitely are there. It doesn't look like you pulled from master during the DAQ workshop.

BenLand100:
How odd - apparently github simply gives the wrong diff when there are conflicts... Should be fixed now unless re-adding my modifications introduced new problems.

It seems it was my multistate branch that was out of date and merging that into this branch brought the old file along with it.

BenLand100:
Interestingly the diff on github now indicates that the unstable branch does not have those changes when it certainly does. Perhaps a github bug?

tlatorre-uchicago:
Yeah it looks like it.

tlatorre-uchicago:
The colors for the Enabled | Disabled on the Fec32 are still green and red, but the boxes are now blue and red.

BenLand100:
These GUIs are a nightmare. Xcode wanted to modify a few more lines and now this conflicts with master again, but if it looks right visually and compiles this Fec32.xib should be preferred to master.

marzece:
merged.

marzece:
I'm going to leave this pull request open and leave unstable as it is for a couple of days because I contacted GitHub about the diff issue. If they don't get back after a week, I'll just close it.

PR #25, opened 2016-01-22
Fix lock behavior for MTC GUI

jcaravaca:
Fix the locks for the 4 different tabs in the MTC GUI. Now it locks when global security is enable and when a run that is not maintenance run is on going. So, for firing pedestals you need to be in maintenance run from now on (enable DAQ expert bit in the bit word).
This has been tested locally and at the test stand already.


tlatorre-uchicago:
Javi, is this independent of the standard runs pull request or does it depend on it for the maintenance run lock?

jcaravaca:
@tlatorre-uchicago, this is independent of the standard runs PR and only the MTC object is changed

tlatorre-uchicago:
merge.

tlatorre-uchicago:
merged.

Issue #27, opened 2016-02-01
XL3 initialization time is too slow


tlatorre-uchicago:
It currently takes about 14 seconds to initialize all the XL3s at run start.

tlatorre-uchicago:
This is partially fixed by pull request #36, but the run actually doesn't wait to start until all XL3s have been initialized.

tlatorre-uchicago:
fixed in #53.

PR #30, opened 2016-02-01
initialize XL3s in parallel when run starts

tlatorre-uchicago:
This pull request should initialize the XL3s in parallel at the start of a run. When the XL3 model receives a notification that the run is about to start it posts a notification which blocks the start of the run and then starts a thread to initialize the XL3. When the initialization is done, it posts another notification releasing the lock on the run start.

This solves Issue #27.


tlatorre-uchicago:
Superseded by pull request #36.

Issue #34, opened 2016-02-07
impossible to set triggers for individual channels


tlatorre-uchicago:
Currently it is impossible to set the N100 and N20 triggers for individual channels. The reason is that triggers are always enabled for online channels with HV on. See https://github.com/snoplus/orca/blob/master/Source/Objects/Custom%20Hardware/SNO%2B/XL3/ORXL3Model.m#L1183. The correct way to do this is to enable triggers based on the FEC32 model variables.

tlatorre-uchicago:
Fixed in pull request #36.

Issue #35, opened 2016-02-09
changing pedestal mask shouldn't affect which channels are read out


tlatorre-uchicago:
Currently if you set the pedestal mask for only some of the slots, the readout for all other slots is turned off.

tlatorre-uchicago:
Fixed in pull request #36.

Issue #33, opened 2016-02-04
XL3 model has hardcoded database reference


tlatorre-uchicago:
See https://github.com/snoplus/orca/blob/master/Source/Objects/Custom%20Hardware/SNO%2B/XL3/ORXL3Model.m#L1023.

tlatorre-uchicago:
Fixed in pull request #36.

Issue #31, opened 2016-02-01
PMT states squares in FEC32 change whether hardware is updated or not


tlatorre-uchicago:
In the PMT States matrix, the square color is updated whether anything is sent to the XL3 or not. Ideally, the lower half of the square should represent the actual hardware settings and so remain unchanged if the crate init fails.

tlatorre-uchicago:
Fixed in pull request #36.

PR #36, opened 2016-02-09
updates to the XL3 code

tlatorre-uchicago:
This updates a lot of the XL3 code, specifically the init crate function which is used to load most hardware settings to the crate. Some of the big changes are:

- switch to using same structure names as ML403 code which makes it easier to keep track of what happens in the XL3 when ORCA sends stuff
- use CouchDB reference from SNOPModel when getting ECALs (no more hardcoded password)
- add a new blocking crate init function (so we can be sure the crate is finished initializing before starting the run)
- fixes a bug in the setTrigger100ns function in the Fec32 model which never actually set the trigger mask for N100
- adds safety checks before run start and prints red warnings to log if it changes hardware
- initialize XL3s in parallel when run is starting

Note that it is now possible within a run to change the sequencer, triggers, and CMOS read settings for individual channels.

This fixes issues #34, #33, #31, and #27.


tlatorre-uchicago:
Also fixes issues #35, #15, and #16.

tlatorre-uchicago:
merged.

PR #39, opened 2016-02-16
Added the wpxmlrpc library and the XmlrpcClient wrapper class for com…

EdLeming:
…munication with the tellie and smellie servers

I this could be seen as equivalent to the hiredis library and complementary RedisClient class made by Eric / Tony.


tlatorre-uchicago:
I just tried to compile this using XCode 5 and I'm getting three errors:

Expected ':' for lines 47, 58, and 70 in WPXMLRPCEncoder.h

To test compilation with XCode 5 you can download it from the Apple developer website.

tlatorre-uchicago:
Note also that there are two lines in ORMJDDewarScaleModel that don't compile on XCode 5.0, but those aren't our problem (I fix them when compiling by just fixing them according to whatever XCode suggests).

EdLeming:
Thanks for catching. I added a macro to resolve.

tlatorre-uchicago:
merged.

Issue #40, opened 2016-02-16
NSLog blocks


marzece:
Maybe everyone already knew this but as far as I can tell [this](https://github.com/snoplus/orca/blob/master/Source/Main%20Dialogs/StatusWindow/ORStatusController.m#L194) chunk of code gets called each time NSLog gets called.
The waitUntilDone:YES means that function will block until the Main Thread is not busy. So any thread that calls NSLog will sit there doing nothing until the Main Thread is free. But it's certainly possible that the Main Thread is waiting for that other (now blocking) thread to do something. So both threads will block basically for eternity.

I discovered this problem by that exact thing happening, except instead of eternity it was 2 seconds (there was a timeout). 

There are two solutions as I see it. 
1.) Change the waitUntilDone option from YES to NO. Then it won't block. But changing how NSLog works will effect pretty much everything in ORCA. 
2.) Make sure NSLog statements in critical threads (such as the ones that receive commands from the XL3) are all launched in yet another separate thread

I'm in favor of talking to Mark Howe and depending on what he says going for option 1. But I thought I'd bring it up here to see if anyone else has thoughts on the matter.

marzece:
I may have spoken too soon. The (oneway void) return type is apparently supposed to keep the function call from blocking, [link](http://stackoverflow.com/questions/5494981/use-case-of-oneway-void-in-objective-c). Nevertheless the call does block and changing the waitUntilDone from YES to NO does seem to fix it. So the oneway void thing just makes things slightly more confusing.

marzece:
I sent Mark Howe an email regarding this. He said

> Please do not change any objects other than SNO objects. The NSLog code is done that way for thread safety. If you are running into a deadlock, you can implement a logging method in your thread that prints via the main thread.

tlatorre-uchicago:
Fixed

PR #41, opened 2016-02-23
Fix NSLog deadlock issue

marzece:
This solves issue #40 . I've implemented some unit tests that (I think) prove that this is a safe change. These unit tests are on my UnitTests branch ([link] (https://github.com/marzece/orca/blob/UnitTests/OrcaTests/NSLogTest.m)). But if you don't want to take the trouble to (git) checkout that branch I'll summarize what the tests say. 
Here is what tests pass/fail before this change.
ortests_before
And here is what tests pass/fail after this change
ortests_after

As you can see the only thing that is different is testDeadlock goes from a fail to pass. 
If you're curious about the other two tests that fail both times it's because NSLog only actually prints when the main loop is allowed to run. Those two tests log from a thread that isn't the main thread, and the output is checked before the main loop is ever given a chance to run. There are tests that do allow the main loop to run (they have the suffix _wRML (with Run Main Loop)) and those pass both before and after my change.

As a side note I'd also like to pull in my unit tests in hopes of providing a spring board for even more tests and therefore better/safer code. But that discussion can be had separately.

### DIFF
--- a/Source/Main Dialogs/StatusWindow/ORStatusController.m
+++ b/Source/Main Dialogs/StatusWindow/ORStatusController.m
@@ -191,7 +191,7 @@ SYNTHESIZE_SINGLETON_FOR_ORCLASS(StatusController);
 
 - (oneway void) printAttributedString:(NSAttributedString*)s1
 {
-    [self performSelectorOnMainThread:@selector(mainThreadPrint:) withObject:s1 waitUntilDone:YES];
+    [self performSelectorOnMainThread:@selector(mainThreadPrint:) withObject:s1 waitUntilDone:[NSThread isMainThread]];
 }
 
 


tlatorre-uchicago:
Woah! Awesome! I definitely think we should check these tests into ORCA. Are the unit tests run in XCode?

marzece:
Yeah, it's supported by xcode 5 and greater

marzece:
I talked with Mark Howe a bit more. He says we can merge this into master whenver we feel satisfied that it's not secretly bad. So I'd say this is good to merge into unstable. I'll use it tomorrow during the testing and hopefully not see any unexpected behavior pop up. Then after that it can be merged into master

tlatorre-uchicago:
merged

PR #42, opened 2016-02-23
Add Unit tests for ORStatusController

marzece:
I created these tests to make sure PR #41 didn't break everything. These tests would increase orca's test coverage from 0.0% to ~0.0%. However, I think they're still valuable b/c it would provide a template for people to add more tests. And that could hopefully lead to a new dawn of rigorously tested and safe code. Anyways I'm eager to know what thoughts others may have.


tlatorre-uchicago:
merged

PR #43, opened 2016-02-25
Implement new panic down as recommeneded by Ben L

marzece:
For the sake of posterity here is the email from @BenLand100 whence these changes take their inspiration
>If you've got the time and motivation, I'm certainly not going to stand in your way. The simplest first pass is to take out the three lines here https://github.com/snoplus/orca/blob/master/Source/Objects/Custom%20Hardware/SNO%2B/XL3/ORXL3Model.m#L3709 and replace with methods to send your new packet to the XL3, then once the panic down is done call https://github.com/snoplus/orca/blob/master/Source/Objects/Custom%20Hardware/SNO%2B/XL3/ORXL3Model.m#L3293 to resync Orca with the XL3 HV state. I can clean up / remove the panic down control logic in my next pull request (which I haven't started, so don't worry about conflicts), but making the suggested changes would effectively bypass the old logic.

I'd like to get this pulled into unstable before tomorrow so that I can use it to test the ML403 code changes that go along with this.

### DIFF
--- a/Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
+++ b/Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
@@ -3702,14 +3702,35 @@ void SwapLongBlock(void* p, int32_t n)
 
 - (void) hvPanicDown
 {
-    [self setIsPollingXl3:NO];	
+    [self setIsPollingXl3:NO];
     if ([self isTriggerON]) {
         [self hvTriggersOFF];
     }
-    [self setHvPanicFlag:YES];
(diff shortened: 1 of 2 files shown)


marzece:
Ah yeah, that very reasonably could be why it had trouble figuring things out after a timeout. ORCA did actually say that the XL3 had disconnected but that disconnect function gets called when the main loop is next available not immediately upon a timeout. So I got both that exception and the disconnect message. Anyways I updated it and I'll test things out tomorrow.

marzece:
This is superseded by #46. So I'll go ahead and close it

PR #46, opened 2016-03-13
HV Monitoring Updates

BenLand100:
Here are the changes discussed in the DAQ workshop. This has been tested on the teststand.

The former 'control' thread now lives as long as an XL3 is connected and reads out the HV status every second. In the event of a >100V discrepancy (arbitrary but the same as the ramp discrepancy check) between the setpoint and the readback an alarm will be posted and the new 'Accept Readback'  button (replaces 'Get Status' since this is now automatic) becomes available while all other HV actions are suspended. If the user presses the 'Accept Readback' button, the readout value becomes the setpoint and HV control resumes. If the high voltage recovers on its own (intermittent readout issue) the monitoring thread realizes and resumes control automatically but the operator must still clear the alarm.  

There is now a convenient spot in the HV control thread to add any checks of HV status one could imagine (current limits have been requested but not yet implemented). 

Also, this includes #43. While the panic down will always work, depending on whether or not the XL3 times out and disconnects, Orca will currently either treat the panic down as a discrepancy requiring user intervention before HV control can be resumed (no disconnect) OR will recover automatically (disconnects).


BenLand100:
I found that there was some duplication of checks in the method to turn the supplies on and off that have now been folded into the HV init thread. As they were there was a possible race condition that could have led to the strange setpoint ramping that @tlatorre-uchicago observed that should now be fixed. This will need re-checking on the test stand.

Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
```diff
@@ -3702,14 +3691,37 @@ - (void) setHVSwitch:(BOOL)aOn forPowerSupply:(unsigned char)sup
 
 - (void) hvPanicDown
 {
-    [self setIsPollingXl3:NO];	
+    [self setIsPollingXl3:NO];
     if ([self isTriggerON]) {
```
marzece: You should be able to get rid of this check. The DoPanicDown command turns triggers off automatically.

BenLand100:
This now includes bugfixes from today's teststand debugging and should be ready for detector testing.

tlatorre-uchicago:
@BenLand100 , does this also include the fixes from your hvmon2 branch now?

tlatorre-uchicago:
Also, could you add a comment above the commented out initialized checks explaining why they aren't needed anymore or just delete them?

BenLand100:
Merged hvmon2 and removed the init code. This is now up-to-date with what we tested this morning.

Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
```diff
@@ -2290,7 +2314,7 @@ - (void) writeXl3Mode
         args->dataAvailMask = swapLong(args->dataAvailMask);
     }
 
-    [self setXl3ModeRunning:YES];
+    //[self setXl3ModeRunning:YES];
```
tlatorre-uchicago: Why is this commented out?

Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
```diff
@@ -2301,7 +2325,7 @@ - (void) writeXl3Mode
                     @"Set XL3 mode failed; error: %@ reason: %@\n",
                     [e name], [e reason]);
     }
-    [self setXl3ModeRunning:NO];
+    //[self setXl3ModeRunning:NO];
```
tlatorre-uchicago: same here.

Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
```diff
@@ -3702,14 +3691,35 @@ - (void) setHVSwitch:(BOOL)aOn forPowerSupply:(unsigned char)sup
...
+    }
+    if(result->errorFlags)
+    {
+        NSLog(@"There was a problem performing panic down. Try again or ramp crate manually");
+        [self safeHvInit];
+        return;
+    }
+    [self safeHvInit];
+    NSLog(@"%@ panic down completed.\n", [[self xl3Link] crateName]);
+    //NSLog(@"%@ panic down completed, hit HV ON to recover\n", [[self xl3Link] crateName]);
```
tlatorre-uchicago: delete this line?

Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
BenLand100: No idea, maybe this is a @marzece question. Ditto below.

tlatorre-uchicago: I'm not sure I understand. Is this something Eric committed?

marzece: Don't look at me. I've never seen that line of code before in my life.

marzece: https://github.com/BenLand100/orca/commit/b20b9c052532f89bf86241cb723edcad93522431 This seems to be the commit where the line was commented out.

tlatorre-uchicago: Ah. Yes that commit was me, but we have since figured this out (well, the solution is don't call changeXl3Mode() from another thread) and so it was uncommented again in a later commit.

Is this just a result of a manual merge resolution?

tlatorre-uchicago: Here is the commit where it is uncommented: https://github.com/snoplus/orca/commit/6a15dd43f2b58048373994ae671eff45e2265aed.

So, the question is, is it showing up as uncommented in this pull request because of an incorrent manual merge or is this a git failed automatic merge from your current HEAD to snoplus:master?

tlatorre-uchicago: Hm. Wait nevermind that's not the commit. I'm confused now. I definitely uncommented them, and they are uncommented in master, but I can't seem to find the right commit.

tlatorre-uchicago: Ah. I see. The commit @marzece pointed to is only on your branch Ben. Not sure how it got in there, but that's not on master.

I do remember committing something like that, but maybe it was just for testing on the teststand and you somehow pushed that to your branch?

BenLand100: I think it was a rider on the panic down branch from @marzece hence the finger pointing. If you all know where it came from and think it should be uncommented I'll switch it back.

tlatorre-uchicago:
merged.

PR #49, opened 2016-04-03
Update function to load MTCA thresholds

tlatorre-uchicago:
This pull request updates the function loadTheMTCADacs in the MTC model to use the MTC server command instead of clocking in the dac values directly. The MTC server can then keep track of the thresholds so that it saves the correct values to the detector state database.


tlatorre-uchicago:
merged.

PR #50, opened 2016-04-03
update shipEPEDRecord

tlatorre-uchicago:
Currently, the shipEPEDRecord sends the EPED record to ORCA which will then put it in the data stream. This pull request updates that function to send the EPED record to the new data stream server so that it eventually gets to the builder.

Note: The current functionality function does *not* send the EPED record to mark subruns.

Also note that the GTID flag is automatically filled in by the MTC server.


tlatorre-uchicago:
merged.

PR #53, opened 2016-04-10
fix a couple issue with run starts.

tlatorre-uchicago:
Previously ORAddRunStateChangeWait notifications would not cause the run start
to pause. We know do hardware initialization after ORRunSecondChanceForWait
notification which does wait until all models have finished initializing.

Also, keep track of when a run start is due to a rollover and don't initialize
any hardware.


tlatorre-uchicago:
merged.

PR #54, opened 2016-04-11
deal with run stops correctly

tlatorre-uchicago:
Currently, the builder flushes all events at the end of *every* run regardless of whether it is a soft or hard run stop.

Now, the builder only flushes events if the next run will not restart, i.e. the user hit the stop run button.


tlatorre-uchicago:
merged.

PR #55, opened 2016-04-12
SNO+ Settings Tab

tlatorre-uchicago:
Updates the SNO+ GUI by renaming the CouchDB tab to Settings tab and adding fields for the hostnames and ports for the MTC, XL3, data, and log servers.


tlatorre-uchicago:
Tested on the teststand April 12, 2016. Going to merge before testing UG so that Javi can continue working on the SNO+ GUI.

tlatorre-uchicago:
merged.

PR #56, opened 2016-04-13
Bug fix for standard run crash

jcaravaca:
The standard run control would set the run type to NULL
when failing contacting orcaDB.
Set it to an empty string if the DB is not reached.


jcaravaca:
@tlatorre-uchicago, this should fix the SNOP model crash when starting a run. Feel free to test it at the test stand or I can do it tomorrow.

PR #57, opened 2016-04-13
SNO+: update from SNO+ settings when object is added.

tlatorre-uchicago:
Previously, when the SNO+ settings tab was updated, the SNO+ model would update
all the relevant models (MTC, CAEN, and XL3), but if an object was later added
the settings would not be in sync.

Now, the MTC, CAEN, and XL3 objects look for a notification to be posted when
objects are added, and if it is themselves, then they update these settings
from the SNO+ model.


tlatorre-uchicago:
merged.

Issue #44, opened 2016-03-09
Switching between two tabs for changing the ped frequency is annoying.


marzece:
A possible solution would be to make a GUI centered around running pedestals.

tlatorre-uchicago:
Fixed by #59.

PR #59, opened 2016-04-17
MTC: Add convenient extra pulser rate field

marzece:
In response to issue #44


tlatorre-uchicago:
This field does not seem to get updated if you change the pulser rate in the Settings tab.

tlatorre-uchicago:
Also fails to compile on XCode 5.1. To fix this you need to manually open up the XIB in a text editor and:

```
s/labelColor/controlTextColor
```

It's a bit of a hack, but that's the only way I've figured out how to get it to compile. Alternatively, adding the new element using XCode 5.1 may work.

tlatorre-uchicago:
merged.

PR #60, opened 2016-04-18
Bug fix when NULL standard run type or version

jcaravaca:
This PR fixes the crash when the SR type or SR version fields retrieved from orcaDB are NULL. It sets a timeout of 2 seconds if CouchDB can't be reached when updating the SR. It also minimizes the print out and cleans up the code.


tlatorre-uchicago:
merged.

PR #61, opened 2016-04-18
SNO+: update 100 ns and 20 ns trigger mask fields in MB struct

tlatorre-uchicago:
Currently, the ML403 uses a bit set in the width and delay fields to enable
or disable the 100 ns and 20 ns triggers. This commit also sets the mask
fields which are used by the detector database. A future update to the ML403
code will use the mask fields and then the code which sets the bits in the
width and delay fields can be deleted.


tlatorre-uchicago:
merged.

PR #62, opened 2016-04-20
Add Run Type Word to SNO+ control

jcaravaca:
Basic performances of the Run Type Word on the SNO+ panel. It's kept sync'ed with main RunControl and checks the mutually exclusive bit part.


jcaravaca:
@tlatorre-uchicago this PR is intended to include a basic runtype word. I've just updated it with today's discussion. It was tested OK at the teststand, so it'd be nice to get this merged asap so that I can keep working on the others features we discussed and want to include.

Source/Experiments/SNOP/SNOPController.m
```diff
@@ -431,59 +394,27 @@ - (IBAction)maintenanceBoxAction:(id)sender {
...
-    //A bit of test code to see a 32-bit word
-    /*NSMutableString *str = [NSMutableString stringWithFormat:@""];
-     for(NSInteger numberCopy = maskValue; numberCopy > 0; numberCopy >>= 1)
-     {
-     // Prepend "0" or "1", depending on the bit
-     [str insertString:((numberCopy & 1) ? @"1" : @"0") atIndex:0];
-     }*/
-}
-
+//- (IBAction)newRunAction:(id)sender {
```
tlatorre-uchicago: Can you explain more about why this method is commented out?

tlatorre-uchicago:
I just tried it out at the teststand. Here are a couple of issues:

- color behind the stop lights is wrong (also, seems to have changed the color of the main run control)
- new run button does not work
- start run button is not greyed out when run is on going
- loadStandardRunToHW is declared in SNOPModel.h but never defined in SNOPModel.m
- run type bits do not match the discussion (or am I missing something?)
- standard runs are not visible if the connection fails on the first attempt (to reproduce have the password wrong when you first boot up, then correct it)

A couple of questions:

- Is it possible to just hardcode the run type? This would make it a lot easier to get ORCA started up and I don't think we expect it to change very often.
- Are we going to stick with the standard run versions?

A couple of issues brought up during testing:

- Button names are confusing (Load Stored Values, Load Default Values, etc.) Load from what to what? Overwrite stored values from where to where?
- New value is a little confusing (really represents the current MTC model state)

Source/Experiments/SNOP/SNOPController.m
jcaravaca: So far I didn't implement any newrunaction and there is no such action in the base clase either. The intention is adding this in another PR. This is only for the run type word and I would like to keep it simple as we now that it works as it is now.

jcaravaca:
This PR is just for the runtype. I was plannign to get into the rest when this is merged, now that we tested it and we know it works.
- I hardcoded the runtype with the definition we came up during that last meeting. Is it not showing to you? Maybe it's loading the txt file automatically? I'll look into that.
- The SR versions will be an expert tool and should be grayed out for users.
- I'll look into the new run, start run and SR GUI in general by separate in another PR.

jcaravaca:
This should fix the fact you don't see the hardcoded labels.
I've removed the declaration of loadStandardRunToHW.
I had a quick look to the issue of the bkg color for the traffic lights and it's a bug introduced in 872c6851f63d723af42d8e7308fd77437cc9dedd, nothing to do with the new SNO+ panel.

tlatorre-uchicago:
merged, but I'll leave this open since it has some comments for the next pull request.

PR #63, opened 2016-04-25
SNO+: add #pragma pack(1) to XL3 DB types also

tlatorre-uchicago:
We really should be packing everything that is sent over a socket. This needs to be coordinated with pull requests to penn_daq, the ML403 code, and the XL3 server.


tlatorre-uchicago:
merged.

PR #64, opened 2016-04-25
SNO+:Update XL3 set sequencer function

marzece:
This makes use of the latest and greatest ML403 firmware which adds a Set Sequencer command. This commit should not be merged until that code is also merged.
https://github.com/pennneutrinos/fullml403/pull/10


tlatorre-uchicago:
merged.

PR #67, opened 2016-05-03
Run start fixes

tlatorre-uchicago:
This pull request updates how run starts/stops are handled. Run starts now do the following:

Cold Run Start
--------------------
- turn triggers off
- initialize CAEN (have to do this first)
- initialize MTC from standard run settings (except trigger mask)
- initialize crates in parallel
- fire SOFT_GT
- suspend MTC readout
- enable triggers
- fire SOFT_GT
- send run header to builder
- resume MTC readout

Continuous Run Start
--------------------------------
- fire SOFT_GT
- suspend MTCD readout
- if not rollover: change MTC settings
- fire SOFT_GT
- send run header
- resume readout

Run Stop
----------------
- fire SOFT_GT
- turn triggers OFF
- wait for XL3 and MTC to read out
- send builder end of run command


tlatorre-uchicago:
This is a work in progress. I still need to make sure that settings aren't changed if it's a diagnostic run.

tlatorre-uchicago:
@jcaravaca , could you have a look at my changes to loadStandardRun and make sure you are OK with them.

tlatorre-uchicago:
On second thought, I don't think we should special case the diagnostic run right now. Now that the crate settings are not reloaded on continuous run starts, we should just make the diagnostic run have thresholds similar to physics run like it was done in SNO.

The reason for this is that if we special case the diagnostic run it makes the GUI design a lot harder. For example, it's not clear what should be shown for the trigger thresholds for diagnostic run. Given that most problems will be channel related, reloading the MTC settings by switching to a diagnostic run should not be a problem if the operator wants to debug an issue with one of the channels.

Thoughts?

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -491,49 +494,266 @@ - (void) runAboutToRollOver:(NSNotification*)aNote
...
+        /* Load the CAEN settings. This should automatically reset it's
+         * counter to zero. */
+        if ([caen initAtRunStart]) {
+            NSLogColor([NSColor redColor], @"failed to load CAEN settings at run start!\n");
+            goto err;
+        }
+
+        /* Load standard run thresholds to the MTC model. Abort the run start
+         * if this fails. */
+        if ([self loadStandardRun:standardRunType withVersion:standardRunVersion]) {
```
jcaravaca: I think we should be loading the DEFAULT when not in expert mode:
`BOOL locked	= [gSecurity isLocked:ORSNOPRunsLockNotification];`
`if(locked) {`
`[self setStandardRunVersion:@"DEFAULT"];`
`    if ([self loadStandardRun:standardRunType withVersion:@"DEFAULT"]){`
`    NSLogColor([NSColor redColor], @"failed to load standard run settings at run start!\n");`
`    goto err;`
`} else{`
`     if ([self loadStandardRun:standardRunType withVersion:standardRunVersion]){`
`     NSLogColor([NSColor redColor], @"failed to load standard run settings at run start!\n");`
`     goto err;`
`     }`
`  }`
`}`

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -491,49 +494,266 @@ - (void) runAboutToRollOver:(NSNotification*)aNote
...
-- (void) runAboutToStart:(NSNotification*)aNote
-{
-    if (rolloverRun) {
-        /* If this is a rollover run, we don't initialize any hardware. */
-        rolloverRun = NO;
-        return;
+    /* Load standard run thresholds to the MTC model. Ideally here we would
+     * check and if the settings haven't changed, we wouldn't need to write
+     * anything, but currently this is not implemented. */
+    if ([self loadStandardRun:standardRunType withVersion:standardRunVersion]) {
```
jcaravaca: Same here

jcaravaca:
As it is now, the loadStandardRun will reload the settings from the DB, so if we tweak the values during a Maintenance run and forget to overwrite into the DB, will rollover to the old values. There are two solutions for this:
- We don't reload for Maintenance runs (I think this would be actually an old SNO diagnostic run): just substitute the condition I comment above by:
`BOOL lockedOrNotRunningMaintenance = [gSecurity runInProgressButNotType:eMaintenanceRunType orIsLocked:ORSNOPRunsLockNotification];`
- The loadStandardRun method detects whether any value has changed wrt the DB and ask the user whether want to abort or change settings in the DB. The problem with this is that will introduce dead time by the user interaction.

tlatorre-uchicago:
Superseded by another commit.

Issue #70, opened 2016-05-04
Print statement: "XL3 error occured while setting squencer" does not have newline at end


fbdescamps:

tlatorre-uchicago:
Fixed by #71.

PR #71, opened 2016-05-04
Add new line to set sequencer related error mesg

marzece:
Solves #70  and bumps my issue solving stats.


tlatorre-uchicago:
merged.

PR #75, opened 2016-05-04
Standard Runs Fixes

jcaravaca:
-- Add button to make sure Standard Runs are properly loaded from DB
-- Include button to load hardcoded safe detector state
-- Add proper functionality to Start/Stop buttons
-- Add Start/Stop custom methods. Will probably change by the new PR handling cold start/rollover/stop runs
-- Add quick links functions
-- Make Maintenance bit mutually exclusive
-- Rename 'Run Version' to 'Test Runs' and make it an expert tool


Source/Experiments/SNOP/SNOPController.m
```diff
@@ -200,9 +217,7 @@ -(void) awakeFromNib
 - (void) registerNotificationObservers
 {
     NSNotificationCenter* notifyCenter = [NSNotificationCenter defaultCenter];
-    NSArray*  objs = [[(ORAppDelegate*)[NSApp delegate] document] collectObjectsOfClass:NSClassFromString(@"ORRunModel")];
-    ORRunModel* theRunControl = [objs objectAtIndex:0];
-    objs = [[(ORAppDelegate*)[NSApp delegate] document] collectObjectsOfClass:NSClassFromString(@"ORMTCModel")];
+    NSArray* objs = [[(ORAppDelegate*)[NSApp delegate] document] collectObjectsOfClass:NSClassFromString(@"ORMTCModel")];
     ORMTCModel* mtcModel = [objs objectAtIndex:0];
```
tlatorre-uchicago: Need to check if objs has at least one element.

Source/Experiments/SNOP/SNOPController.m
```diff
@@ -327,130 +339,103 @@ -(IBAction)fireTellie:(id)sender
...
-    [self refreshStandardRunVersions];
+    NSString* standardRun = [model standardRunType];
+    if([standardRunPopupMenu indexOfItemWithObjectValue:standardRun] == NSNotFound){
+        NSLogColor([NSColor redColor],@"Standard Run \"%@\" does not exist in DB. \n",standardRun);
+    }
+    else{
+        [standardRunPopupMenu selectItemWithObjectValue:standardRun];
+    }
     
+    if([[model standardRunType] isEqualToString:@"HIGH THRESHOLDS"]) {
```
tlatorre-uchicago: I don't think we need a standard run named high thresholds. I was just imagining that the button would be used to set high thresholds if the database can't be reached.

jcaravaca:
OK, this should be ready to go. I've include your comments but two things: the popup when you open the experiment to allow connection to the DB and the error querying the DB at ORCA start. Probably both related because I don't see any of those locally. I can look into those during the week at the teststand but I don't think any of the items are critical.
A caveat is that this PR will change the structure of the .Orca file leaving the old ones deprecated. We'll need to comment out 314 of SNOPModel.m, run ORCA, save the .Orca file, uncomment that line and run ORCA again normally.

tlatorre-uchicago:
Can you give me more detailed info on how to fix this ECA type thing? I keep getting uncaught exceptions in _runDocumentWorker no matter what I do.

tlatorre-uchicago:
merged.

PR #80, opened 2016-05-05
Added shipSubRunRecord to SNOPModel.m.

EdLeming:
This will allow subRun boundaries to be sent onto the builder by calling [runControl prepareForNewSubRun]. EPEDrecords are currently used to flag subrun boundaries at the builder. Here we re-purpose the pedestalWidth variable of the EPED record to hold the subRunNumber.

**I need to speak with Phil about how this will be handelled at the builder end. Is this how he expects this to be defined?


Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -664,6 +666,41 @@ - (void) updateEPEDStructWithNSlopePoint: (unsigned long) nTSlopePoints
...
+     *     #define EPED_FLAG_SUBRUN 0x1000000
+     *
+     * All fields associated with EPED settings are set to zero. The mtc server
+     * adds a GTID value to the record before piping it down to the builder.
+     *
+     */
+
+    //get the run controller
+    NSArray*  objs3 = [[(ORAppDelegate*)[NSApp delegate] document] collectObjectsOfClass:NSClassFromString(@"ORRunModel")];
+    ORRunModel* runControl = [objs3 objectAtIndex:0];
```
EdLeming: need to check here that the length of the array is greater than 0 otherwise this will be an out of bounds exception

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -627,9 +627,11 @@ - (void) runStateChanged:(NSNotification*)aNote
 
 - (void) subRunStarted:(NSNotification*)aNote
 {
-    //EPED record
+    //Ship subrunrecord - Just a special case of an eped record
+    [self shipSubRunRecord];
+    
     //TRIG record?
```
tlatorre-uchicago: we should ask Phil whether we should ship a trig record. If not, we can delete this line

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -627,9 +627,11 @@ - (void) runStateChanged:(NSNotification*)aNote
...
-    //update orcadb run document
+    //update orcadb run document?
```
tlatorre-uchicago: the orca db run document is already deprecated, so I don't think we need to add any info here. You can delete this comment.

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -627,9 +627,11 @@ - (void) runStateChanged:(NSNotification*)aNote
 
 - (void) subRunStarted:(NSNotification*)aNote
 {
```
tlatorre-uchicago: Can you add a quick comment here describing the notion of sub runs in ROC

tlatorre-uchicago:
merged.

Issue #68, opened 2016-05-04
Flipping HV on/off state in SNO+ control GUI


fbdescamps:
When starting ORCA and  turning on HV on a single crate, we noticed that the ```PMTs HV is ON/OFF``` button was flipping (on off on off on off, etc).

jcaravaca:
Is that the global control button? Was only that or the PANIC DOWN being grayed out as well?

fbdescamps:
Both were in flipping state, I think it was like this:
 * HV ON was combined with enabled panic down
 * HV OFF was combined with disabled panic down

But the HV was ON.

fbdescamps:
![hvflipping 001](https://cloud.githubusercontent.com/assets/1840217/15002684/938f3f38-1158-11e6-8b9a-1075063f69e9.png)

jcaravaca:
Does the HVMaster panel look like normal? Indicates ON/OFF the correct crates?

fbdescamps:
Ah, we did not check this. Also, for some unknown reason, the flipping stopped and it went to ```HV ON`` when it indeed was.

tlatorre-uchicago:
It's flipping again, and the HV master tab is stable with just crate 6
showing HV ON.

BenLand100:
https://github.com/snoplus/orca/blob/master/Source/Experiments/SNOP/SNOPController.m#L460
In the event that there is a value in ```aNote``` the ```globalHVON``` variable is not set as an OR of XL3 states. This notification is shipped with a value for ```aNote``` for each connected XL3 so ```globalHVON``` should be set by looping over the XL3s outside that ```if``` statement and all will be good.

Issue #77, opened 2016-05-05
HV Threading Crash


tlatorre-uchicago:
When starting ORCA, it immediately crashed with the following error:

```
2016-05-04 20:34:27.097 Orca[15183:10673501] An uncaught exception was raised
2016-05-04 20:34:27.097 Orca[15183:10673501] forwardInvocation:: NSUndoManager 0x28d7f550 received forwarded invocation while invocation target is nil. Call prepareWithInvocationTarget: before invoking setHvBVoltageDACSetValue:
2016-05-04 20:34:27.099 Orca[15183:10673501] (
	0   CoreFoundation                      0x9127b343 __raiseError + 195
	1   libobjc.A.dylib                     0x92d43a2a objc_exception_throw + 276
	2   CoreFoundation                      0x9127b25d +[NSException raise:format:] + 141
	3   Foundation                          0x90aee124 -[NSUndoManager _forwardTargetInvocation:] + 169
	4   Foundation                          0x90aee069 -[NSUndoManagerProxy forwardInvocation:] + 43
	5   CoreFoundation                      0x911b7122 ___forwarding___ + 482
	6   CoreFoundation                      0x911b6f1e _CF_forwarding_prep_0 + 14
	7   Orca                                0x008493f0 -[ORXL3Model setHvBVoltageDACSetValue:] + 112
	8   Orca                                0x00865a04 -[ORXL3Model(private) _hvInit] + 1556
	9   Foundation                          0x90ab2dcf -[NSThread main] + 45
	10  Foundation                          0x90ab2d21 __NSThread__main__ + 1538
	11  libsystem_pthread.dylib             0x915e3ecf _pthread_body + 138
	12  libsystem_pthread.dylib             0x915e3e45 _pthread_body + 0
	13  libsystem_pthread.dylib             0x915e1f0e thread_start + 34
)
2016-05-04 20:34:27.099 Orca[15183:10673501] *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'forwardInvocation:: NSUndoManager 0x28d7f550 received forwarded invocation while invocation target is nil. Call prepareWithInvocationTarget: before invoking setHvBVoltageDACSetValue:'
*** Call stack at first throw:
(
	0   CoreFoundation                      0x9127b343 __raiseError + 195
	1   libobjc.A.dylib                     0x92d43a2a objc_exception_throw + 276
	2   CoreFoundation                      0x9127b25d +[NSException raise:format:] + 141
[...]

PR #87, opened 2016-05-06
HV Threading Crash Fix

BenLand100:
This removes the undo manager from a few high voltage related methods that are called by the init or monitoring threads. The undo manager introduced a race condition leading to a rare crash observed in #77. Also includes a minor change - hv related error messages are now red in the Orca log.

Pull request from Benjamin Land


Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
```diff
@@ -2317,7 +2313,7 @@ - (void) writeXl3Mode
         args->dataAvailMask = swapLong(args->dataAvailMask);
     }
 
-    //[self setXl3ModeRunning:YES];
+    [self setXl3ModeRunning:YES];
```
BenLand100: @tlatorre-uchicago these lines are commented out in unstable but not in master, which is probably why my last PR had them commented out. I'm not sure how that happened, but I believe we decided uncommented was correct?

BenLand100:
@tlatorre-uchicago this also now posts notifications to main thread.

BenLand100:
@tlatorre-uchicago besides removing the undo manager calls that resulted in #77, it's conceivable that wrapping all the ORXL3ModelHvStatusChanged notifications in main thread dispatchers might also fix #68. We'd probably need the whole detector to check either because if I recall neither was seen on the teststand, but these should only be cosmetic changes as far as hardware settings are concerned.

tlatorre-uchicago:
Thanks! merged.

PR #92, opened 2016-05-07
Better handle multiple threads waiting on XL3 response.

### LINKED ISSUE #69: It takes 60 seconds for the HV thread to re-appear after an ML403 reset button press
That's pretty long...

BenLand100:
Threads waiting for a packet now exit when the link is marked as disconnected. Additionally they are counted and a reconnect can only happen after all waiting threads have exited. This should regulate the reconnect process and fix #69.

### DIFF
--- a/Source/Objects/Custom Hardware/SNO+/XL3/XL3_Link.m
+++ b/Source/Objects/Custom Hardware/SNO+/XL3/XL3_Link.m
@@ -55,6 +55,7 @@ NSString* XL3_LinkAutoConnectChanged    = @"XL3_LinkAutoConnectChanged";
 
 @implementation XL3_Link
 @synthesize fifoTimeStamp = _fifoTimeStamp,
+pendingThreads,
 readFifoFlag = _readFifoFlag;
 
 - (id) init
@@ -64,6 +65,8 @@ readFifoFlag = _readFifoFlag;
 	commandSocketLock = [[NSLock alloc] init];
 	coreSocketLock = [[NSLock alloc] init];
 	cmdArrayLock = [[NSLock alloc] init];
+    connectionLock = [[NSLock alloc] init];
+    pendingThreads = 0;
 	[self setNeedToSwap];
 	connectState = kDisconnected;
 	cmdArray = [[NSMutableArray alloc] init];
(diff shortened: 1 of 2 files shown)


Source/Objects/Custom Hardware/SNO+/XL3/XL3_Link.m
```diff
@@ -595,31 +610,47 @@ - (void) readXL3Packet:(XL3Packet*) aPacket withPacketType:(uint8_t) packetType
...
+        [self setPendingThreads:pendingThreads-1];
+        [connectionLock unlock];
+        [foundCmds release];
+        foundCmds = nil;
+        @throw [NSException exceptionWithName:@"ReadXL3Packet XL3 disconnected"
+            reason:[NSString stringWithFormat:@"XL3 disconnected for %@ <%@> port: %lu\n", [self crateName], IPNumber, portNumber]
+            userInfo:nil];
+    } else if ([foundCmds count] > 1) {
+        //FIXME: ask tony about this case
+		NSLogColor([NSColor redColor],@"Multiple responses for XL3 command with packet type: %d and packet num: %d from %@ <%@> port: %d\n",
```
BenLand100: @tlatorre-uchicago when multiple responses are found, this message was printed and the first response was returned. Does that seem reasonable still? I made it red since presumably it's an error.

BenLand100:
This should probably be validated by testing the high voltage response to a ml403 reset.

Source/Objects/Custom Hardware/SNO+/XL3/XL3_Link.m
tannerbk: I think this is fine. Theoretically this should never happen since the XL3 server queues requests to a single XL3, so making its red error is good.

tlatorre-uchicago:
The test stand is free right now if you want to give it a quick test. I think we should probably test this on the full detector sometime before Monday because any error is more likely to show up with lots of crates.

Quick question: What is the purpose of the pendingThreads variable in addition to the connection lock. Wouldn't just having the lock be acquired and released in readXL3Packet and then acquiring and releasing the lock in connectToPort accomplish the same thing? Or is this so that you could potentially allow more than 2 threads to call readXl3Packet?

BenLand100:
pendingThreads is for multiple threads to work at the same time, yeah. The existing code seemed to have been written to support that, and there is the hv thread, polling thread, and main thread that might send packets and expect response.

I'll be back in 1-2 hours and could call in, but if you want to test it without me it might just work.

BenLand100:
I amended this to remove that annoying debug.

tlatorre-uchicago:
merged.

PR #93, opened 2016-05-08
SMELLIE

EdLeming:
Changes to allow stable SMELLIE running for the May FTB dark running.


EdLeming:
@tlatorre-uchicago, this should be ready to go. There's a line in ORMJDDewarScaleModel that won't compile on 5.1.1, but I guess this is probably a known issue.

marzece:
It looks like you'll need to make this work in the post force-push world that we're living in. The easiest way to do so (that I know of) is to squash all of your commits into a single commit (which is advised anyway) then from a branch that is current with snoplus master git cherry-pick that squashed commit. Then push that to github.

tlatorre-uchicago:
Ed, did you fix this after testing last night and verify that it was fixed?

EdLeming:
As far as I could tell the ellie code worked as expected last night.

I think the issues we saw derive from one of the links in the data stream
chain, which was being overloaded bye 1kHz of peds from a few thousand
channels

tlatorre-uchicago:
I thought things went haywire? How do you know it was a data stream problem?

tlatorre-uchicago:
Also, which part of the code is sending pedestals?

tlatorre-uchicago:
Also, it appears that the builder was getting MTC trigger masks with all triggers masked in. This I don't think could be a data stream problem, unless there is a serious bug which is sending random bytes to the builder.

EdLeming:
To start the builder spit out a load of orphans, then the GTID in xsnoed went crazy. In the end it was resolved by re-initing the xl3s. Maybe some buffers got overloaded somewhere. 

Currently smellie slave mode requests peds from the MTCD. Eventually this will be replaced by some communication with Tubii, but that is not finished yet

EdLeming:
Yeah the code doesn't mask triggers in, they had just been all masked in for stability running by Ian. We didn't think to change them for the first run - which is when things went crazy. When we dropped to 5 crates (the ones to be used for SMELLIE) it settled down.

EdLeming:
...sorry, I'm talking about masking crates in, not the trigger mask on the MTCD. That I have no idea about.

tlatorre-uchicago:
So, did you update the code to only mask in the TUB and SMELLIE?

EdLeming:
I don't mask in any triggers at all. That will have to be done by the operator.

tlatorre-uchicago:
But don't you only want the pedestals to go to SMELLIE?

EdLeming:
That's right, we just sent them to all the crates last night through operator error

tlatorre-uchicago:
I think it would be nice to just code that in. If you start a SMELLIE run then it only masks in the TUB and SMELLIE and then resets the crate mask at the end of the run.

EdLeming:
OK, I guess SMELLIE will be in one of the 'spare' slots shown on the mtc gui?

tlatorre-uchicago:
merged.

PR #107, opened 2016-05-10
Mtca thresholds fix

tlatorre-uchicago:
This pull request fixes the ordering of the MTCA thresholds.


tlatorre-uchicago:
merged.

PR #108, opened 2016-05-10
fix the bug in the SNO+ controller which was causing the PMT HV to blink on and off

tlatorre-uchicago:
As Ben pointed out in issue #68, the flipping HV state is caused by the globalHVON variable not being set properly when a notification is sent by a single XL3.

This pull request fixes the issue by keeping track of which power supplies are on in the SNO+ controller using a bitmask, where 1 means the HV is on and 0 means it is off. The first 18 bits represent the power supply A in the crates, and bit 19 represents the state of power supply B in crate 16.

### DIFF
--- a/Source/Experiments/SNOP/SNOPController.m
+++ b/Source/Experiments/SNOP/SNOPController.m
@@ -51,6 +51,9 @@ snopGreenColor;
 -(id)init
 {
     self = [super initWithWindowNibName:@"SNOP"];
+
+    hvMask = 0;
+
     return self;
 }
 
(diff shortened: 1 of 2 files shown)


Source/Experiments/SNOP/SNOPController.m
```diff
@@ -453,10 +456,7 @@ - (void) dbDebugDBIPChanged:(NSNotification*)aNote
 }
 
 - (void) hvStatusChanged:(NSNotification*)aNote
-{ dispatch_async(dispatch_get_main_queue(), ^{
```
jcaravaca: Can you keep this line unless there is something critical about it? This might cause problems if updates the GUI in a secondary thread

tlatorre-uchicago: I've verified that this notification is only posted on the main thread. There are lots of spots in the code where updates to the GUI happen assuming the main thread is running. The recommended way of fixing this is to always post these notifications on the main thread.

BenLand100:
Was merged.

Issue #105, opened 2016-05-10
save triggers off state


tlatorre-uchicago:
Orca should save the triggers OFF/ON state in the .Orca file, otherwise triggers are turned off every time you restart Orca.

tlatorre-uchicago:
Fixed in #109.

PR #109, opened 2016-05-10
save triggers OFF/ON state to .Orca file

tlatorre-uchicago:
Fixes issue #105.

### DIFF
--- a/Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
+++ b/Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
@@ -1435,8 +1435,8 @@ void SwapLongBlock(void* p, int32_t n)
         [self setHvNominalVoltageB:2500];
     }
 
-    for (i=0; i<12; i++) {
-        [self setXl3VltThreshold:i withValue:[decoder decodeFloatForKey:[NSString stringWithFormat:@"ORXL3ModelVltThreshold%hd", i]]];
+    for (i = 0; i < 12; i++) {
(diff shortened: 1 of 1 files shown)


Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
```diff
@@ -1415,6 +1419,13 @@ - (id)initWithCoder:(NSCoder*)decoder
     [self setHvNominalVoltageA: [decoder decodeIntForKey:@"ORXL3ModelHvNominalVoltageA"]];
     [self setHvNominalVoltageB: [decoder decodeIntForKey:@"ORXL3ModelHvNominalVoltageB"]];
     [self setXl3Mode:           [decoder decodeIntForKey:@"Xl3Mode"]];
+    [self setIsTriggerON:       [decoder decodeBoolForKey:@"isTriggerON"]];
```
jcaravaca: Are we sure this is the real state of the triggers? It seems to me that this is a model variable updated only when clicking the trigger buttons. Is it warrantied to be in sync with the XL3s?

tlatorre-uchicago: No, this is not the state of the triggers. It represents a software switch that disables or enables you to switch triggers on. It is not really in sync with the XL3s. The way it works is this:

- if triggers are OFF then you cannot switch triggers on (even from the PMT states box)
- if triggers are ON then you can switch triggers on (but it doesn't guarantee that triggers are on)

so basically it's just a handy way to ensure that triggers aren't on while ramping down.

BenLand100:
Was merged.

PR #112, opened 2016-05-11
disable triggers before loading trigger thresholds at run start

tlatorre-uchicago:
Noticed that changing the MTCA trigger thresholds results in a burst of events, so we should disable triggers while we are loading these thresholds at run start.


marzece:
Wouldn't it be better just to mask out physics triggers? It doesn't make much of a difference really, but also it doesn't make much sense to lose your pulse gts or whatever just b/c you're changing thresholds.

tlatorre-uchicago:
I think you're right, but the time the triggers are off is so small you can't even hear it on the TUB. It would be possible to do this, but I'm not sure it's worth the effort, because it requires a lot more bookkepping, and this only happens on run boundaries.

BenLand100:
Was merged.

PR #113, opened 2016-05-11
SNO+: add resync run functionality

tlatorre-uchicago:
Add the ability to do a resync run, which does a hard stop and start without the user having to hit stop run and then start run. Doing this resets the GTID, which resyncs crate 9 after it goes out of sync :).


BenLand100:
Was merged.

PR #114, opened 2016-05-11
Do not update XL3 HV setpoints on init.

BenLand100:


BenLand100:
Only merge this if we need to use Orca immediately. #115 is the final fix.

BenLand100:
No longer needed.

PR #115, opened 2016-05-11
SNO+: Remove automatic HV setpoint adjustment, enforce HV target maximum

BenLand100:
This updates only Orca's internal setpoint based on the HV readback but will not set it higher than the target. No XL3 setpoint adjustment is done without explicit user intervention. This means the first change in HV on the Orca side will implicitly include this offset. 

This needs loadbox testing.


BenLand100:
Has been tested on detector, ready to merge.

BenLand100:
Was merged.

Issue #85, opened 2016-05-06
Use of blue and red


klabe:
In the XL3 HV control window, the boxes for the relays are coded with red for on and blue for off.  In the Pmt States box, the status is coded as red for off and blue for on.  Given that these are often used at the same time, this is ridiculously confusing.

tlatorre-uchicago:
Fixed in #123.

Issue #29, opened 2016-02-01
Delete detector wide buttons from XL3 GUI


tlatorre-uchicago:
At the bottom of the XL3 GUI there are three detector wide buttons:

- panic down
- triggers on
- triggers off

Although the panic down is not as confusing, the triggers are very easy to mistake for a local change to that specific crate.

fbdescamps:
This should be partly replaced by the global always-on-top detector-wide box, which will have the 'emergency' buttons: 

 * big red 'PANIC DOWN'
 * big yellow/orange 'TRIGGER OFF'

The detector-wide triggers on is in the danger zone of the SNO+ control panel run control.

tlatorre-uchicago:
@fbdescamps , Now that we have these buttons in the SNO+ GUI, what do you think about removing them from the XL3 GUI?

fbdescamps:
Hmm.. so, I think we can delete them from the XL3 GUI, if we are absolutely sure that it works from the SNO+ GUI.

tlatorre-uchicago:
Fixed in #123.

PR #123, opened 2016-05-18
SNO+: Switch red/blue on HV relays. Remove detector wide buttons from 3 GUI.

BenLand100:
Note: Actions were also removed for detector wide buttons in the XL3_LinkController.
I confirmed that the main SNO+ GUI does not use these actions for detector wide
trigger on/off or HV panic down, so this shouldn't break anything.

Fixes issues #85 and #29 - superficial changes shouldn't need testing.


tlatorre-uchicago:
merged.

Issue #84, opened 2016-05-06
normal mode on run start


klabe:
Crate 0 did not enter normal mode on run start.

marzece:
We saw something like this happen with crates 9 & 10 earlier. Basically when we turned the crates on and initted them most crates were in normal mode after that. However that was not the case for 9 and 10. 
Here's the shift report (though I've already given more info here than I did in the report) 
http://snopl.us/shift/view/30aca1acdff6411fa9c8bd6225c51fff

fbdescamps:
Is there any update on this one? @tlatorre-uchicago ?

tlatorre-uchicago:
I haven't been able to reproduce this. I suspect it's a non issue. I don't know exactly what caused the earlier problems but here are my suspicions:

1. If the XL3 window drop down selection was set to NORMAL mode but never actually sent, that would cause ORCA to think that the crate was in NORMAL mode (I think?) which means that it wouldn't try to set it to normal mode on run start.

2. The issue Eric reported seems to be unrelated to run starts. The XL3 is not put into normal mode by default after a crate init, so that is expected.

I'm going to leave this issue up though just as a reminder that it might be a good idea to set the crate to normal mode after loading the settings and not at run starts, since I think that is when most people naturally expect it to happen.

tlatorre-uchicago:
Pull request #254 sets the XL3s to normal mode after loading the hardware settings.

PR #125, opened 2016-05-20
XL3 will not init if HV is on for crate.  Cannot XL3 init w/ xilinx unless XL3 is in init mode.

pershint:


Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
```diff
@@ -1865,6 +1865,20 @@ - (int) initCrate: (MB *) mbs xilinxLoad: (int) xilinxLoad shiftRegOnly: (uint32
...
 
+    /*Check XL3 is in init mode before init-ing with xilinx. */
+    [self updateXl3Mode];
+    if ([self xl3Mode] != INIT_MODE && xilinxLoad) {
+        NSLogColor([NSColor redColor], @"XL3 %02d not in init mode.  Aborting initialization.\n", [self crateNumber]);
+        goto err;
+    }
+
+    /*Check that HV Relays for XL3 are open. */
+    [self readHVStatus];
```
tlatorre-uchicago: This method is probably not the one you want, since this updates the HV readback values, not the switch states.

tlatorre-uchicago:
@pershint Any update on this? I think the plan was to ask @BenLand100 whether you needed to explicitly update the switch model variables before checking or not.

pershint:
@tlatorre-uchicago Sorry for the slow response; I followed up with @BenLand100 on this and he suggested explicitly updating the switch model variables, but to test the code on the test stand to make sure there's no errors related to race conditions.  I will test this on the test stand next week and push the changes if everything goes well.

tlatorre-uchicago:
Sounds good. I agree it's good to test it on the teststand, but I doubt you are going to catch a race condition in that way. Race conditions are generally rare since they depend on thread execution timing which you may be unlikely to see during normal circumstances. If a race condition with the HV code is a possibility, then I think that should be handled by making sure all appropriate variables are locked.

BenLand100:
In this case it would be a benign race condition in setting the boolean switch states. If another thread requests the switch states from the XL3 at nearly the same time it is not guaranteed which result will be set last, and locking would not change the ordering ambiguity. I wouldn't expect this race to cause any issues though.

tlatorre-uchicago:
@BenLand100 are the switch states updated on connection and every time the buttons are pressed? If so, there should be no need to update it again here right?

@pershint I think it's OK to take out the `[self updateXl3Mode]` call. It's true that we have a bug which seems to be leading to an incorrect knowledge of the XL3 state (see Issue https://github.com/snoplus/orca/issues/84), but that's a separate issue and thinking about it now, I don't think we should be correcting for this every place in the code by updating before checking.

BenLand100:
@tlatorre-uchicago that is correct, but the hv switch state is not monitored by Orca so a third party program could potentially switch on a supply that was off and Orca would not know until a user clicked around or the supply was ramped high enough to trigger a setpoint discrepancy. As we don't have any third party HV control programs this should not be an issue, and the HV control logic would need changing to support external programs anyway.

tlatorre-uchicago:
@pershint, Is this ready to merge?

pershint:
@tlatorre-uchicago, it is.  I just tested it on the test stand and everything worked aside from the new line characters (just edited).

Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
```diff
@@ -1865,6 +1865,18 @@ - (int) initCrate: (MB *) mbs xilinxLoad: (int) xilinxLoad shiftRegOnly: (uint32
...
     CrateInitArgs *crateInitArgs;
 
+    /*Check XL3 is in init mode before init-ing with xilinx. */
+    if ([self xl3Mode] != INIT_MODE && xilinxLoad) {
+        NSLogColor([NSColor redColor], @"XL3 %02d not in init mode.  Aborting initialization.\n", [self crateNumber]);
+        goto err;
+    }
+
+    /*Check that HV Relays for XL3 are open. */
+    if ([self hvASwitch] || [self hvBSwitch]) {
```
tlatorre-uchicago: Here you need to check that it is actually doing a full crate init also. The HV gets reset if you do a crate init with shiftRegOnly == 0 because the XL3 then does a crate reset. So, this line should be:

```C
if ((shiftRegOnly == 0) && ([self hvASwitch] || [self hvBSwitch])) {
```

otherwise this will prevent ORCA from changing *any* settings while HV is on.

pershint:
@tlatorre-uchicago I've pushed a new commit with your requested changes.

Just so I understand, it's all right to run an init with high voltage on so long as it is a shift registers only init?

tlatorre-uchicago:
Yeah, unfortunately the names here are a bit weird because of how the XL3 interface was designed. In reality, the crate init function is responsible for doing *both* real crate inits (where the crate is reset and all new values are loaded) and loading new hardware settings (no crate reset).

What happens depends on the shiftRegOnly variable:

shiftRegOnly == 0
--------------------------

- full crate initialization (causes HV to drop to zero)
- all settings are reloaded

shiftRegOnly == 1
--------------------------

- loads CMOS registers (trigger settings)

shiftRegOnly == 2
-------------------------

- loads CMOS registers (trigger settings)
- loads all DACs (voltage thresholds, discriminator thresholds, etc)

tlatorre-uchicago:
merged.

PR #126, opened 2016-05-23
Standard Run Fixes

jcaravaca:
Improvements and bug fixes for the Standard Runs and ECA panels.
Send notifications to main thread.
Do not store detector state bits in run type word for the Standard Runs.
Delete estop buttons in hvMaster.

### DIFF
--- a/Source/Experiments/SNOP/SNOPController.m
+++ b/Source/Experiments/SNOP/SNOPController.m
@@ -45,6 +45,8 @@ smellieRunFile,
 snopBlueColor,
 snopRedColor,
 snopOrangeColor,
+snopBlackColor,
+snopGrayColor,
 snopGreenColor;
 
 #pragma mark ¥¥¥Initialization
@@ -206,7 +208,11 @@ snopGreenColor;
     [self setSnopRedColor:[NSColor colorWithSRGBRed:255./255. green:102./255. blue:102./255. alpha:1]];
     [self setSnopGreenColor:[NSColor colorWithSRGBRed:0./255. green:150./255. blue:0./255. alpha:1]];
     [self setSnopOrangeColor:[NSColor colorWithSRGBRed:255./255. green:178./255. blue:102./255. alpha:1]];
+    [self setSnopBlackColor:[NSColor colorWithSRGBRed:0./255. green:0./255. blue:0./255. alpha:1]];
+    [self setSnopGrayColor:[NSColor colorWithSRGBRed:0./255. green:0./255. blue:0./255. alpha:0.5]];
 
+    //Update conection settings
+    [self updateSettings:nil];
     //Sync runnumber with main RunControl
     [self updateRunInfo:nil];
     [self findRunControl:nil];
@@ -215,10 +221,14 @@ snopGreenColor;
     [self mtcDataBaseChanged:nil];
     //Update runtype word
     [self runTypeWordChanged:nil];
-    //Refresh SRs
-    [self refreshStandardRuns:nil];
-    //Update conection settings
-    [self updateSettings:nil];
+    //Lock Standard Runs menus at init
+    if([model standardRunType] == nil){
+        [standardRunPopupMenu setEnabled:false];
+        [standardRunVersionPopupMenu setEnabled:false];
+    }
+    else{
+        [self refreshStandardRunsAction:nil];
+    }
     //Pull the information from the SMELLIE DB
     [model getSmellieRunListInfo];
     [super awakeFromNib];
@@ -320,7 +330,6 @@ snopGreenColor;
     [self runStatusChanged:nil];
     [self SRTypeChanged:nil];
     [self SRVersionChanged:nil];
-    [model setIsEmergencyStopEnabled:TRUE]; //enable the emergency stop
     [self runsLockChanged:nil];
     [self runsECAChanged:nil];
     [self runTypeWordChanged:nil];
@@ -351,8 +360,9 @@ snopGreenColor;
 {
     
     NSString* standardRun = [model standardRunType];
-    if([standardRunPopupMenu indexOfItemWithObjectValue:standardRun] == NSNotFound){
+    if(standardRun != nil && ![standardRun isEqualToString:@""] && [standardRunPopupMenu indexOfItemWithObjectValue:standardRun] == NSNotFound){
         NSLogColor([NSColor redColor],@"Standard Run \"%@\" does not exist in DB. \n",standardRun);
+        return;
     }
     else{
         [standardRunPopupMenu selectItemWithObjectValue:standardRun];
@@ -360,16 +370,20 @@ snopGreenColor;
     
     if([[model standardRunType] isEqualToString:@"HIGH THRESHOLDS"]) {
         [model setStandardRunVersion:@"DEFAULT"];
-    } else {
+    }
+    else if(![standardRun isEqualToString:@""] && standardRun != nil){
         [self refreshStandardRunVersions];
     }
+
 }
 
 -(void) SRVersionChanged:(NSNotification*)aNote
 {
+    
     NSString* standardRunVersion = [model standardRunVersion];
-    if([standardRunVersionPopupMenu indexOfItemWithObjectValue:standardRunVersion] == NSNotFound){
+    if(standardRunVersion != nil && ![standardRunVersion isEqualToString:@""] && [standardRunVersionPopupMenu indexOfItemWithObjectValue:standardRunVersion] == NSNotFound){
         NSLogColor([NSColor redColor],@"Standard Run Version \"%@\" does not exist in DB. \n",standardRunVersion);
+        return;
     }
     else{
         [standardRunVersionPopupMenu selectItemWithObjectValue:standardRunVersion];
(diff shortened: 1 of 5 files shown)


jcaravaca:
@tlatorre-uchicago this should be ready to go now

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -1457,8 +1395,8 @@ - (void)encodeWithCoder:(NSCoder*)encoder
     [encoder encodeInt:[self logPort] forKey:@"logPort"];
 
     //Runs tab
-    [encoder encodeObject:[self standardRunType] forKey:@"SNOPStandardRunType"];
-    [encoder encodeObject:[self standardRunVersion] forKey:@"SNOPStandardRunVersion"];
+    //[encoder encodeObject:[self standardRunType] forKey:@"SNOPStandardRunType"];
+    //[encoder encodeObject:[self standardRunVersion] forKey:@"SNOPStandardRunVersion"];
```
tlatorre-uchicago: Why are these lines commented out? Is the plan not to save the standard run settings to the Orca file?

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -185,8 +183,10 @@ - (void) setMTCPort: (int) port
         [caen setMTCPort:port];
     }
 
-    [[NSNotificationCenter defaultCenter]
-     postNotificationName:@"SNOPSettingsChanged" object:self];
+    dispatch_async(dispatch_get_main_queue(), ^{
```
tlatorre-uchicago: Why did you change all of the setters to run on the main thread?

Source/Experiments/SNOP/SNOPController.m
```diff
@@ -1630,12 +1638,23 @@ -(void) displayThresholdsFromDB {
...
-        NSLogColor([NSColor redColor],@"Standard Run Version not set \n");
+        NSLogColor([NSColor redColor],@"Standard Run not set (ignore this if you just restarted ORCA) \n");
+        return;
+    }
+    //If no version: display null values and quit
+    if([model standardRunVersion] == nil || [[model standardRunVersion] isEqualToString:@""]){
+        for (int i=0; i<[standardRunThresStoredValues numberOfRows];i++) {
+            [[standardRunThresStoredValues cellAtRow:i column:0] setStringValue:@"--"];
+        }
+        NSLogColor([NSColor redColor],@"Test Run not set (ignore this if you just restarted ORCA) \n");
```
tlatorre-uchicago: I think showing red log messages on every start up isn't a good idea since it will lead to alarm fatigue, i.e. operators will get used to red alerts whenever starting ORCA. What about just not loading or displaying thresholds until the user clicks the load values from DB button?

tlatorre-uchicago:
@jcaravaca I've added a few comments. After addressing these, I'll test it out on the teststand.

Source/Experiments/SNOP/SNOPModel.m
jcaravaca: That'd be neat, but then it conflicts with the fact that we want to reload the SR only when clicking the reload button. To be more specific: now when you restart ORCA, it doesn't contact the SR RB automatically, but you need to press 'reload'. In this case, it doesn't make sense to save those fields in the .ORCA file.

Source/Experiments/SNOP/SNOPModel.m
jcaravaca: Some setters send notifications to change the GUI and it is recommended to send them to the main thread. That might be what was causing the SNOP GUI to freeze. The XL3 object does the same thing.

Source/Experiments/SNOP/SNOPModel.m
tlatorre-uchicago: Yeah, it looks like a lot of methods in the XL3 Link controller do that, but I think we should instead force code running in separate threads to call these methods on the main function instead of putting it in the setter itself, otherwise we are going to incur the overhead of putting these on the queue for every setter method and the fact that it is called as dispatch_async() means that the variable actually isn't guaranteed to be set when you reach the next line after calling the setter.

Source/Experiments/SNOP/SNOPModel.m
tlatorre-uchicago: Yeah, it looks like a lot of methods in the XL3 Link controller do that, but I think we should instead force code running in separate threads to call these methods on the main function instead of putting it in the setter itself, otherwise we are going to incur the overhead of putting these on the queue for every setter method and the fact that it is called as dispatch_async() means that the variable actually isn't guaranteed to be set when you reach the next line after calling the setter.

Source/Experiments/SNOP/SNOPController.m
jcaravaca: This is a bit tricky: this function is called by SRVersionChanged, which is called when that SR field has changed OR the SNOP window is opened. If no SR field name is selected this warning pops up and the display will show dashes rather than the numbers from the DB.
This needs to be like this because not only the user can change the SR field (e.g. RunScripts do as well), so I think the logic needs to be kept like this.
On the other hand, I could be more explicit on the message and show it in black color, as technically it might not be a warning:
'Standard Run not set: reload the DB and select a SR in the dropdown menu'

tlatorre-uchicago: In that case, it's not an error at all right? I think in that case, no log message is needed. Can you elaborate in the comments the difference between nil and the empty string? It's not clear what the two mean.

jcaravaca:
@tlatorre-uchicago I've implemented the minimal changes we were talking about last time. I think it's worth to merge this if there are no major issues to unlock the GUI development.

Source/Experiments/SNOP/SNOPModel.m
tlatorre-uchicago: This dispatch_async still needs to be removed.

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -225,8 +225,10 @@ - (void) setMTCHost: (NSString *) host
         [caen setMTCHost:host];
     }
 
-    [[NSNotificationCenter defaultCenter]
-     postNotificationName:@"SNOPSettingsChanged" object:self];
+    dispatch_async(dispatch_get_main_queue(), ^{
```
tlatorre-uchicago: here too.

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -243,8 +245,9 @@ - (void) setXL3Port: (int) port
     [xl3_server disconnect];
     [xl3_server setPort:port];
 
-    [[NSNotificationCenter defaultCenter]
-     postNotificationName:@"SNOPSettingsChanged" object:self];
+    dispatch_async(dispatch_get_main_queue(), ^{
```
tlatorre-uchicago: here too.

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -275,8 +278,10 @@ - (void) setXL3Host: (NSString *) host
         [[xl3 xl3Link] disconnectSocket];
     }
 
-    [[NSNotificationCenter defaultCenter]
-     postNotificationName:@"SNOPSettingsChanged" object:self];
+    dispatch_async(dispatch_get_main_queue(), ^{
```
tlatorre-uchicago: here

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -648,7 +650,10 @@ - (void) runInitialization:(NSNotification*)aNote
                                 @"", @"Reason",
                                 @"", @"Details",
                                 nil];
-    [[NSNotificationCenter defaultCenter] postNotificationName:ORAddRunStartupAbort object: self userInfo: userInfo];
+    dispatch_async(dispatch_get_main_queue(), ^{
```
tlatorre-uchicago: here

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -703,7 +708,9 @@ - (void) runAboutToStart:(NSNotification*)aNote
...
```
tlatorre-uchicago: here

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -752,7 +759,9 @@ - (void) runStarted:(NSNotification*)aNote
                                 @"", @"Reason",
                                 @"", @"Details",
                                 nil];
-    [[NSNotificationCenter defaultCenter] postNotificationName:ORRequestRunHalt object: self userInfo: userInfo];
+    dispatch_async(dispatch_get_main_queue(), ^{
```
tlatorre-uchicago: here

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -784,8 +793,9 @@ - (void) runAboutToStop:(NSNotification*)aNote
         NSDictionary *userInfo = [NSDictionary dictionaryWithObjectsAndKeys:
                                   @"waiting for MTC/XL3/CAEN data", @"Reason",
                                   nil];
-
-        [[NSNotificationCenter defaultCenter] postNotificationName:ORAddRunStateChangeWait object: self userInfo: userInfo];
+        dispatch_async(dispatch_get_main_queue(), ^{
```
tlatorre-uchicago: here

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -1082,7 +1094,10 @@ - (void) setOrcaDBIPAddress:(NSString*)orcaIPAddress
 		[[NSUserDefaults standardUserDefaults] setObject:self.orcaDBConnectionHistory forKey:[NSString stringWithFormat:@"orca.%@.orcaDBConnectionHistory",[self className]]];
 		[[NSUserDefaults standardUserDefaults] setInteger:self.orcaDBIPNumberIndex forKey:[NSString stringWithFormat:@"orca.%@.orcaDBIPNumberIndex",[self className]]];
 		[[NSUserDefaults standardUserDefaults] synchronize];
-		[[NSNotificationCenter defaultCenter] postNotificationName:ORSNOPModelOrcaDBIPAddressChanged object:self];
+        dispatch_async(dispatch_get_main_queue(), ^{
```
tlatorre-uchicago: Not sure where this is called from, but I think running this on the main thread should be done separately if anyone is trying to change it outside of the main thread, and not in the setter.

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -1113,7 +1128,9 @@ - (void) setDebugDBIPAddress:(NSString*)debugIPAddress
 		[[NSUserDefaults standardUserDefaults] setObject:self.debugDBConnectionHistory forKey:[NSString stringWithFormat:@"orca.%@.debugDBConnectionHistory",[self className]]];
 		[[NSUserDefaults standardUserDefaults] setInteger:self.debugDBIPNumberIndex forKey:[NSString stringWithFormat:@"orca.%@.debugDBIPNumberIndex",[self className]]];
 		[[NSUserDefaults standardUserDefaults] synchronize];
-		[[NSNotificationCenter defaultCenter] postNotificationName:ORSNOPModelDebugDBIPAddressChanged object:self];
+        dispatch_async(dispatch_get_main_queue(), ^{
```
tlatorre-uchicago: same here

tlatorre-uchicago:
@jcaravaca I've added comments for lines where you've added asynchronous dispatches to the main thread that I think need to be removed. If you think I missed something and these are correct changes, let me know. Otherwise, after they are removed I'll test on site and merge.

jcaravaca:
I think the 'dispatchs' came back as an artifact of the rebase with master. Now they should be gone except the one in '_waitForBuffers'.

tlatorre-uchicago:
merged.

PR #134, opened 2016-07-15
Fix eca type causing errors when given weird values

### LINKED ISSUE #128: Crashes occured during run starts and save operation
I suspect the issue(s) that we saw were primarily caused by the fact that we had dragged a new FEC from the catalog into crate 18 (which wasn't on). Since it never got values from an ECAL document or a crate init or anything like that the card had a number of un-initialized values that perhaps were just random chunks of memory. So when we tried to do a run start/save the .orca file it tried to write those values to disk. And that caused a seg-fault.  Perhaps related is that while trying to debug that issue I had set a break point and I noticed that at the breakpoint the snopmodel's pointer would flip quickly between NULL and a reasonable value. I suspect that is caused by a thread that wasn't affected by the breakpoint alloc-ing and dealloc-ing the snopmodel.

marzece:
This kind of fixes #128. That error was caused by the fact that ECA_type in the snopmodel was for some reason getting set to really weird values. ECA_Type is an NSString* but the SNOTMain.orca file used UG had that ECA_Type encoded as an NSNumber* with a value of 75. And since no one recently had opened the SNOPModel and changed it to a regular value that weird value has persisted for a long time. If you look at the stack trace referenced in #128 you can see that it tries to cast the NSNumber to an NSString. 

So the actual fix was to change the code so it doesn't attempt to decode that value and then just hardcode in a normal value ( "ECA_type" = @"PDST") then save that into the .orca file. 
Then get rid of the hardcoding and make everything normal again.

Since I have no clue ho ECA_type = NSNumber* with value of 75 it's not unthinkable that it will happen again.  These changes should make it so if somehow ECA_type does get a really weird value orca doesn't crash all the time. I'm unsure how orca should handle that really but I think crashing is probably not the correct way.

### DIFF
--- a/Source/Experiments/SNOP/SNOPController.m
+++ b/Source/Experiments/SNOP/SNOPController.m
@@ -1204,9 +1204,14 @@ snopGreenColor;
 {
     
     //Refresh values in GUI to match the model
-    NSInteger* index = [model ECA_pattern] -1;
+    int index = [model ECA_pattern] -1;
+    if(index < 0)
+    {
+        NSLogColor([NSColor redColor], @"ECA bad index returned\n");
+        return;
+    }
     [ECApatternPopUpButton selectItemAtIndex:index];
-    [ECAtypePopUpButton selectItemWithTitle:[[model ECA_type] retain]];
+    [ECAtypePopUpButton selectItemWithTitle:[model ECA_type]];
     int integ = [model ECA_tslope_pattern];
     [TSlopePatternTextField setIntValue:integ];
     integ = [model ECA_nevents];
--- a/Source/Experiments/SNOP/SNOPModel.m
+++ b/Source/Experiments/SNOP/SNOPModel.m
@@ -1768,13 +1768,18 @@ err:
 
 - (NSString*)ECA_type
 {
-    return ECA_type;
+    return [NSString stringWithFormat:@"%@",ECA_type];
 }
 
 - (void) setECA_type:(NSString*)aValue
 {
-    ECA_type = aValue;
-    [[NSNotificationCenter defaultCenter] postNotificationName:ORSNOPModelRunsECAChangedNotification object:self];
+    if(aValue != ECA_type)
+    {
+        NSString* temp = ECA_type;
+        ECA_type = [aValue retain];
+        [temp release];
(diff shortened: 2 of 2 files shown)


marzece:
Ah I've come up with a good theory for how that weird value got into the .orca file. If you look at  [here](https://github.com/snoplus/orca/commit/b748c3332ddd959683e8e2c32a67c9ecb3f947f3#diff-d902f06cbecbed7872e75d3f3ac05069L133) the eca_type variable was changed from an int to an NSString. However no one ever told the snot_main.orca that the  value should no longer be an int. 

So it's unlikely for this bug to crop up again now that I've fixed the snot_main.orca file to have a string for that variable rather than an int. So I'm not sure how necessary this commit is.

tlatorre-uchicago:
Good catch. I remember this happening to me when testing the pull request, and I had changed the .Orca file when testing, but I guess it never got propagated underground.

I guess in the future we should probably:

1. have a single master copy of the .Orca file
2. Never change the type of these values (if necessary, just use a different name)

marzece:
Yeah I think the only thing you can do to prevent this is to be aware/vigilant. Or switch from relying on the .orca file to using the DB.

 I think this PR still has merit though b/c some of the memory management stuff that was there previously seemed pretty weird and maybe have been added hastily to make the change from int to NSString*. This makes the memory management a bit more standard.

Source/Experiments/SNOP/SNOPController.m
```diff
@@ -1204,9 +1204,14 @@ - (void) runsECAChanged:(NSNotification*)aNotification
 {
     
     //Refresh values in GUI to match the model
-    NSInteger* index = [model ECA_pattern] -1;
```
tlatorre-uchicago: So this was just the wrong type all along. In fact, shouldn't this have essentially caused a memory error every time since it appears to be casting an int to a pointer?

marzece: Yeah, I'm not sure how this gets handled. Maybe you don't get errors so long as you don't try to write to that memory. And then selectItemAtIndex treats the variable gently enough that you can get away with such a thing.

tlatorre-uchicago:
merged.

Issue #132, opened 2016-07-11
DB Url request sometimes fails mysteriously in SNOPController


tannerbk:
This is Eric not Tanner, he left himself logged into github on operug.

   The issue occurs around line [1950 of  SNOPController.m](https://github.com/snoplus/orca/blob/master/Source/Experiments/SNOP/SNOPController.m#L1951).
Basically sometimes the NSError variable is set by the call to sendSynchronousRequest and sometimes it's not set. When an error does occur the output simply says the format of the data is not valid.

Below is a (slightly edited) version of the code. I added extra debug statements. The output of which is also below.

```
    NSLog(@"self = %@\n",self);
    //Now query DB and fetch the SRs
    NSString* urlString = [NSString stringWithFormat:@"http://%@:%@@%@:%u/orca/_design/standardRuns/_view/getStandardRuns",[model orcaDBUserName],[model orcaDBPassword],[model orcaDBIPAddress],[model orcaDBPort]];
    NSString* link = [urlString stringByAddingPercentEscapesUsingEncoding:NSUTF8StringEncoding];
    NSLog(@"link : %@\n",link);
    NSURLRequest* request = [NSURLRequest requestWithURL:[NSURL URLWithString:link] cachePolicy:0/*NSURLRequestReloadIgnoringCacheData*/ timeoutInterval:1];
    NSURLResponse* response=nil;
    NSError* error=nil;
    NSData* data= [NSURLConnection sendSynchronousRequest:request returningResponse:&response error:&error];
    NSString *ret = [[[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding] autorelease];

    NSDictionary *standardRunTypes = [NSJSONSerialization JSONObjectWithData:[ret dataUsingEncoding:NSUTF8StringEncoding] options:0 error:&error];

    if(error) {
        NSLogColor([NSColor redColor], @"here\n errror: %@\n data: %@\n",error,data);
        [model setStandardRunType:@""];
        return;
    }
```

 self = <SNOPController: 0x2a432c00>
16-07-11 12:54:08 link : http://USER_REDACTED:PASSWORD_REDACTED@couch.snopl.us:80/orca/_design/standardRuns/_view/getStandardRuns
16-07-11 12:54:09 here
 errror: Error Domain=NSCocoaErrorDomain Code=3840 "The data couldn’t be read because it isn’t in the correct format." (No value.) UserInfo=0x35603550 {NSDebugDescription=No value.}
 data: (null)

[...]

PR #136, opened 2016-07-18
standard runs database error

tlatorre-uchicago:
This pull request updates the SNO+ controller class to print the error message if the database request fails when fetching the standard runs or standard run versions.

This will help in diagnosing issue #132.


tlatorre-uchicago:
merged.

Issue #130, opened 2016-06-30
CMOS polling output should be unsigned


marzece:
Currently you can get output like the following when a very high rate shows up. 
Why you might get rates that high is a whole other question. 

ch24-31:      298       298       298       298       298       298       298       299 
slot: 13
ch00-07:      299       299       299       299       299       299       299       299 
ch08-15:      299 -2147483648       299 -2147483648       299 -2147483648       299 -2147483648 
ch16-23:      299 -2147483648       299 -2147483648       299 -2147483648       299 -2147483648 
ch24-31:      299 -2147483648       299 -2147483648       299 -2147483648       299 -2147483648

tlatorre-uchicago:
I suspect this is because it doesn't handle the CMOS counter rolling over correctly.

fbdescamps:
@marzece What is the status here?

marzece:
I haven't touched this since my last attempt failed. Since CMOS polling might soon be moved out of orca this may become a non-issue.

fbdescamps:
I am leaving this open until we have the ORCA-independent CMOS polling.

tlatorre-uchicago:
I wonder if the reason your pull request didn't work is because the counters are actually 31 bits instead of 32 bits. I was trying to figure out the size of the register and came across this: https://github.com/pennneutrinos/fullml403/blob/8cb8b8336b8ba474485000e85043ff35882a2970/src/FECUtils.c#L250 which suggests that it's only 31 bits.

@marzece Am I reading that correctly? I couldn't find any other docs which explicitly said how many bits there are in the count register.

marzece:
Huh, whaddya know. From the ML403 code for getting cmos counts
```
readResults = doRW(0x0, CMOS_INTERN_TOTAL(j) + FEC_SEL*i + READ_REG);
errorBit = (readResults & 0x80000000) ? 1 : 0;
...
readResults &= 0x7FFFFFFF;
results->count[numSlots*32+j] = readResults;
```
It would seem it was only 31 bits. That very likely did cause a lot of the issues I ran into with this.

PR #149, opened 2016-08-18
SNO+: Account for roll over in cmos rate polling

marzece:
Addresses issue #130. I haven't been able to test this yet so I'll add a comment when I do. I'm just opening this PR so that no one else spends time on this issue.

### DIFF
--- a/Source/Objects/Custom Hardware/SNO+/Fec32/ORFec32Model.m
+++ b/Source/Objects/Custom Hardware/SNO+/Fec32/ORFec32Model.m
@@ -1783,8 +1783,10 @@ static unsigned long cratePedMask;  // crates that need their pedestals set
 			if (calcRates) { //only good CMOS count reads get here
                     if(cmosCount[ch] > theCount){
                         // theCount rolled over
+                        uint32_t dumbRate = (theCount - cmosCount[ch]) * sampleFreq;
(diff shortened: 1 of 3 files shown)


Source/Objects/Custom Hardware/SNO+/Fec32/ORFec32Model.m
```diff
@@ -1779,9 +1779,16 @@ - (BOOL) processCMOSCounts:(unsigned long*)rates calcRates:(BOOL)aCalcRates with
     unsigned char ch;
     for (ch=0; ch<32; ch++) {
         if (aChannelMask & 1UL << ch) {
-            theCount = rates[ch];
+            theCount = counts[ch];
 			if (calcRates) { //only good CMOS count reads get here
-					theRate = (theCount - cmosCount[ch]) * sampleFreq;
+                    if(cmosCount[ch] > theCount){
+                        // theCount rolled over
+                        theCount = (ULONG_MAX - cmosCount[ch]) + theCount;
```
tlatorre-uchicago: @marzece Are you sure the CMOS counters are the same length as an unsigned long? May be best to hardcode the limit since a long may be machine dependent.

marzece:
I updated this to deal with the system dependent memory stuff that @tlatorre-uchicago mentioned. Still haven't tested that it solves the issue without introducing new problems though.

PR #157, opened 2016-09-09
SNO+: Accept PMTIC IDs returned from XL3 on init

marzece:
This is in anticipation of a complementary change being made to the ML403 firmware.


tlatorre-uchicago:
merged.

PR #159, opened 2016-11-21
Implement expert and operator mode and other additions

jcaravaca:
- Proper locking behavior in every element
- Remove test run from display
- Add individual panic down buttons in HVMaster
- Prompt for user confirmation for important actions


tlatorre-uchicago:
Let me know when you would like me to merge this. As a reminder here's a checklist of things to check before merging:

- Make sure it compiles on XCode 5.1.1 (in practice, I check with XCode 4.3 which is installed on the teststand)
- Run Product -> Analyze in XCode to check for any memory leaks (best to run in XCode 8 if you can)
- double check that no commented out code was added (without a clear comment above)
- run tests on teststand and document those tests in a comment here on the pull request

jcaravaca:
This contains some minor changes in the HV logic to handle enabling/disabling buttons according to the lock status. I'd need approval from @BenLand100 and probably we would want to run some HV tests.

jcaravaca:
- Compiles in XCode5.1 (teststand)
- Run 'Analyze' in XCode6.2: no potential memory leaks related with SNO+ objects
- Tested at the teststand:
-- Checked lock status in different scenarios: lock closed/open, run stoped/started, maintenance/physics. Everything looking as expected.
-- Checked the SR display is correct.
-- Didn't test HV buttons since they are Crate-specific. Need to be tested at the detector.

Once this is reviewed and ready for merge I can squash in a single commit.

BenLand100:
The changes to the HV were to the locked states of the buttons so this seems fine. We should do a final HV test once the this and the alarm code is tested and merged in. Since the changes bypass the ```updateHVButtons``` method we'll need to double check that during various edge cases the gui doesn't end up in a weird state but this won't affect the back end of actually changing or monitoring HV.

jcaravaca:
The last commit include functionality of loading the SR settings into HW automatically when the expert user click the 'Load SR' button. This is handy for Maintenance mode, since now we don't need to load settings and then manually load into HW from the MTC object to test our changes.

jcaravaca:
Last commit implements fixes for the 2 issues found at the teststand testing:
- Bug on the load SR button
- Panic down crate 16A/B should be a single button
This should be now ready to go.

tlatorre-uchicago:
merged.

PR #161, opened 2016-11-29
SNO+: Add HV alarm subsystem to XL3 code.

BenLand100:
Uses the current postgres object to send alarm queries for HV monitoring. Currently implements current and voltage limits, current dropout detection, and setpoint discrepancies. Parameters are hardcoded per crate pending full DB integration.

The hardcoded values (ctrl+f TEMPBLOCK) represent previous behavior for the ramps with:
* Explicit hardcoding of nominal values (from twiki) - no longer read from .Orca file
* Setpoint tolerance at 50V
* Over-voltage alarm at NOMINAL+50V
* Over-current alarm at 1000mA - totally arbitrary will need setting later
* Under-current alarm requiring at least 10mA when >500V

The idea is to switch to reading these parameters from the database (code is in place and included in this PR) once the postgres object supports returning query results.

* Compiled on ancient xcode
* Checked for memory leaks
* Didn't modify gui files
* Tested on teststand 11/23
* Tested on detector 11/28

### DIFF
--- a/Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.h
+++ b/Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.h
@@ -108,6 +108,8 @@ enum {
     BOOL hvBSwitch;
     BOOL hvARamping;
     BOOL hvBRamping;
+    BOOL hvAFromDB;
+    BOOL hvBFromDB;
     BOOL hvEverUpdated;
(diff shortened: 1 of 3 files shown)


BenLand100:
@tlatorre-uchicago I snipped the methods that accessed the parameters in the DB so this should be good to go. I'll test the DB access code soon and open a new PR with that.

tlatorre-uchicago:
merged.

PR #162, opened 2016-11-29
Postgres run number

tlatorre-uchicago:
This pull request updates ORCA so that it fetches the run number from the database. It also fixes a few bugs in the ORPQModel.

Tested on the teststand on Monday, November 28 2016.

At the start of a run, the SNO+ model contacts the database to get the next run number. If the database is not connected it aborts the run. If the database is available but the query fails for some reason, the run number is set to 0xffffffff. This run number will then be interpreted by the builder as a "default" run in which it writes events to default_[something].zdab. This means that if the database is unavailable we can still switch runs and get events in XSNOED, but these runs will never make it through data flow which is what we want since if the database is unavailable, the runs will not be considered golden anyways.

### DIFF
--- a/Source/Experiments/SNOP/SNOPModel.m
+++ b/Source/Experiments/SNOP/SNOPModel.m
@@ -651,9 +651,10 @@ resync;
     } else {
         /* If there is no database object, just continue with the existing run
          * number saved to Orca. */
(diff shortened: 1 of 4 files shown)


tlatorre-uchicago:
merged.

PR #163, opened 2016-11-29
	Implemented FetchRow... methods in ORPQResult

boardhead:
Also now syncs orca to current detector state before a hardware wizard operation.
Tested on the test stand and it seems to work well.


tlatorre-uchicago:
merged.

PR #164, opened 2016-12-01
Reset crate

tlatorre-uchicago:
This pull request updates ORCA to send the new crate reset packet. The crate GUI is updated to now have just three buttons: "Reset Crate", "Fetch ECAL Settings to GUI", and "Load Hardware".


tlatorre-uchicago:
merged.

PR #169, opened 2016-12-02
updates the XL3 model to use the MULTI_SET_CRATE_PEDS_ID command

tlatorre-uchicago:
This pull request updates the XL3 model to use the MULTI_SET_CRATE_PEDS_ID command which sets a different pedestal mask for each slot. Previously, the XL3 model was doing low level writing to registers which is harder for the XL3 server to keep track of.


tlatorre-uchicago:
merged.

PR #171, opened 2016-12-03
Update TUBii trigger mask to include ELLIE triggers

icoulter:
Also add ability to choose whether a trigger goes to sync or async output on MTCD as requested by SMELLIE/Armin.


tlatorre-uchicago:
I'm getting the following warnings:

- lots of TUBII xib warnings
- "Property 'trigMask' requires method 'setTrigMask:' to be defined."
- "Instance method '-setTrigMask:syncMask:' not found"

icoulter:
Rejigged things to fix the warnings and also fix something that was left out of the match hardware option.
I don't see any xib warnings anymore (I only saw one) but Eric suggests it could be computer dependent.

tlatorre-uchicago:
merged.

PR #172, opened 2016-12-05
SNO+: Add float retrieval to postgres object and HV parameter retrieval from DB.

BenLand100:
The HV monitoring and control thread now uses the current postgres object to retrieve control and alarm parameters on program start. Without DB access neither control nor monitoring of HV will be enabled as the thread responsible for those actions will not start.

I've tested this locally but not yet on the teststand. If anyone is feeling courageous feel free to give it a try, but I'll get to it when I get to it. The hvparams table is setup on both the detector (dbug) and test (minard) databases with appropriate values including current limits from Noel.

### DIFF
--- a/Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
+++ b/Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
@@ -108,6 +108,8 @@ extern NSString* ORSNOPRequestHVStatus;
 - (void) _post_heartbeat:(int)crate;
 - (void) _trigger_edge_alarm:(int)alarmid;
 - (void) _update_level_alarm:(int)alarmid level:(bool)state;
+- (void)_hv_a_dbparams:(ORPQResult*)result;
+- (void)_hv_b_dbparams:(ORPQResult*)result;
 @end
 
 @implementation ORXL3Model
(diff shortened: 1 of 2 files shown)


BenLand100:
@boardhead I added float and double retrieval to your psql code. Shouldn't be any surprises but you might want to take a quick look.

tlatorre-uchicago:
There are two potential errors when fetching things from the database:

1. The database object hasn't been added to ORCA in which case `[ORPQModel getCurrent]` should return `nil`
2. The database command fails (don't expect this to happen often, but it could happen while testing on the teststand if someone wiped out the table).

What do you think about adding the following behavior:

1. If the database object can't be found, pop up a warning dialog to the user saying that they need to add the PostgreSQL object to the experiment, and then just use the hardcoded nominal values.

2. If the database operation fails pop up a warning dialog which says that the database operation failed and then give the operator the option to use the hardcoded values or try again.

Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
```diff
@@ -4532,9 +4534,39 @@ - (void) _pollXl3
...
-float nominals[] = {2110.0, 2240.0, 2075.0, 2160.0, 2043.0, 2170.0, 2170.0, 2170.0,
-                    2060.0, 2435.0, 2240.0, 2370.0, 2220.0, 2270.0, 1970.0, 2025.0,
-                    1995.0, 1945.0, 2010.0, 2000.0}; //crates 0-19 supply a
+//Callback for database access to hv params for supply a
+- (void)_hv_a_dbparams:(ORPQResult*)result
+{
+    if (result && !hvAFromDB) { // Only do this once
+        [self setHvAFromDB:true];
+        NSDictionary* dict = [result fetchRowAsDictionary];
+        [self setHvNominalVoltageA:[(NSNumber*)[dict valueForKey:@"nominal"] floatValue]];
```
BenLand100: My understanding is that the cast is basically meaningless in objective-c except for clarity, but I happen to know from Phil's code that all the results are wrapped in NSNumber

BenLand100:
Currently I am 'silently' failing if the DB can't be contacted with the caveat that the ORCA heartbeats also won't be issued (since the HV thread never starts), but those two changes would certainly make it more robust in the case that we need HV control and the DB is down for some reason. I'll implement that soon.

To be clear, as written it will keep trying in the event of some problem so if one starts orca without the psql object the user can add it and HV should start once it gets a valid response from the DB.

tlatorre-uchicago:
I think you should also add constraints to the hvparams table to make sure that the columns are not null since that isn't checked for here.

BenLand100:
Superseded by #197

PR #173, opened 2016-12-05
Attempt to fix SNO+ GUI freeze problem

### LINKED ISSUE #73: SNO+ GUI became unresponsive
At some point during running today, the SNO+ GUI became completely unresponsive, and we had to restart it.

jcaravaca:
Fix #73 

Update GUI in the main thread. This is common practice and always recommended: http://stackoverflow.com/questions/25168847/mac-application-window-stops-updating

### DIFF
--- a/Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
+++ b/Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
@@ -3571,7 +3571,7 @@ err:
         [self setHvBNeedsUserIntervention:false];
     }
     @try {
-        [self setHVDacA:[self hvBVoltageDACSetValue] dacB:[self hvBVoltageDACSetValue]];
+        [self setHVDacA:[self hvAVoltageDACSetValue] dacB:[self hvBVoltageDACSetValue]];
     }
     @catch (NSException *exception) {
         NSLogColor([NSColor redColor],@"%@ HV failed to set HV!\n", [[self xl3Link] crateName]);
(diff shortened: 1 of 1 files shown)


Source/Experiments/SNOP/SNOPController.m
```diff
@@ -754,7 +770,8 @@ - (IBAction)panicDownSingleCrateAction:(id)sender {
     
 }
 
-- (IBAction)updatexl3Mode:(id)sender{
+- (IBAction)updatexl3Mode:(id)sender
+{ dispatch_async(dispatch_get_main_queue(), ^{
```
tlatorre-uchicago: I think this is redundant because the this is only ever called in response to a GUI event which is guaranteed to be on the main thread.

tlatorre-uchicago:
Were you able to track down a separate thread calling one of these setters?

I probably wasn't very clear earlier about the problem. The reason I mentioned it before was because it's not always what you want. Here is a specific example:

```
[self setXL33Host:"buffer1"]
# The XL3s look for this notification and set their local hostname when this setter is called
# but if the notification is posted asynchronously, then they will not have updated their variables
# when this next line is run
[xl3 doSomething]
```

In general, when you call a setter you expect that the variable will be set and that all objects will receive the notification.

So, it is fine to post notifications or run other code asynchronously if all it is doing is updating the GUI, but not OK to make all setters post their notifications asynchronously.

jcaravaca:
I never see the freeze out in my laptop so I couldn't track down the problems by checking out the different threads. I'm afraid it will need to be debugged in the detector.
I understand your point, so the setters do not run async now. I just implemented this for the main GUI changes like Standard Runs, HVMaster and Settings. I couldn't do it for the RunControl since that'd imply to touch the core ORCA. If the problem persists I'll contact Mark.

jcaravaca:
Tested at the teststand. If @tlatorre-uchicago don't have any other comment this should be ready to go.

tlatorre-uchicago:
Can you explain how these fixes fix the problem? I just had a quick look through for example the two function calls `displayThresholdsFromDB` and `mtcDatabaseChanged` and I was unable to find these functions being called from any other thread.

There also seem to be other non freeze related changes like those in the standardRunPopupAction and runTypeWordAction method. Are those just bug fixes? If so, could you open a separate pull request?

You've also removed methods related to the detector view in the SNO+ GUI but didn't change the GUI at all. Eric currently has a pull request open to delete most of this, so I think you can delete those changes from this pull request.

jcaravaca:
If a notification is posted in a background thread, then it's my responsibility to send the GUI changes to the main thread to prevent the freeze-out(http://stackoverflow.com/questions/15813764/posting-nsnotification-on-the-main-thread). That's the case of the hvStatusChanged:(NSNotification*) for instance, where the XL3s might send the Notification in a background thread.
In general any GUI change should happen in the main thread so I use dispatch_async just in case. I don't see any downside on that other than ensuring 100% the changes happen in the main thread.

I'll remove the pieces you requested and open a separated PR with the bug fixes.

jcaravaca:
A comment from Mark H.:

>  I have seen that problem (rarely) also. Never been able to catch it with the debugger, tho.  If you can help solve the problem I would be grateful. Whatever the cause, it seems to be just the Run Control dialog -- I have never seen it elsewhere. I also suspected that the GUI was being updated from some other thread. That may be it, but I have not been able to track it down so far. A global 'freeze' of that dialog could also mean that all of its notifications got 'de-registered' somehow. Don't know how that could happen tho.

Issue #167, opened 2016-12-02
orca crash due to uncaught exception in _hvInit


tlatorre-uchicago:
After stopping and starting the XL3 server when ORCA reconnected to the XL3 server there was an uncaught exception in _hvInit. I've attached the debugger output:

```
2016-12-01 16:59:03.627 Orca[49900:303] Invalid color System, labelColor (warning given only once)
2016-12-01 16:59:03.811 Orca[49900:303] 10.4-style NSDateFormatter method called on a 10.0-style formatter, which doesn't work. Break on _NSDateFormatter_Log_New_Methods_On_Old_Formatters to debug. This message will only be logged once.
2016-12-01 16:59:04.020 Orca[49900:303] Could not find image named 'CameraPic'.
2016-12-01 17:01:55.451 Orca[49900:ef2b] ERROR: unable to get the receiver data from the DB!
2016-12-01 17:01:58.502 Orca[49900:1f453] ERROR: unable to get the receiver data from the DB!
2016-12-01 20:08:27.232 Orca[49900:1bf2b] An uncaught exception was raised
2016-12-01 20:08:27.232 Orca[49900:1bf2b] forwardInvocation:: NSUndoManager 0x2a244200 received forwarded invocation while invocation target is nil. Call prepareWithInvocationTarget: before invoking setHvNominalVoltageB:
2016-12-01 20:08:27.234 Orca[49900:1bf2b] (
	0   CoreFoundation                      0x907f0471 __raiseError + 193
	1   libobjc.A.dylib                     0x9904b091 objc_exception_throw + 162
	2   CoreFoundation                      0x907f038b +[NSException raise:format:] + 139
	3   Foundation                          0x92c7546f -[NSUndoManager _forwardTargetInvocation:] + 163
	4   Foundation                          0x92c753c0 -[NSUndoManagerProxy forwardInvocation:] + 43
	5   CoreFoundation                      0x907402f9 ___forwarding___ + 457
	6   CoreFoundation                      0x9074010e _CF_forwarding_prep_0 + 14
	7   Orca                                0x0088873f -[ORXL3Model setHvNominalVoltageB:] + 191
	8   Orca                                0x008a4bc1 -[ORXL3Model(private) _hvInit] + 801
	9   Foundation                          0x92c3916e -[NSThread main] + 45
	10  Foundation                          0x92c390c6 __NSThread__main__ + 1426
	11  libsystem_pthread.dylib             0x933085fb _pthread_body + 144
[...]

PR #174, opened 2016-12-06
SNO+: Fix crash from calling UndoManager methods from non-main thread.

BenLand100:
Removes the ability to undo setting the nominal voltage (we shouldn't do this anyway!) fixing #167


tlatorre-uchicago:
Sweet! Merged.

PR #176, opened 2016-12-06
Fix bug in setting a run time limit.

marzece:
Previously the run time limit field might interpret a number with a
comma in it (e.g. 9,999) as a string and then convert the comma to an
period (9,999->9.999) meaning you couldn't have times in excess of 1000
seconds.

This commit fixes that by change the run time limit NSTextField to a
NSTextField with NumberFormatter. The NumberFormatter is not customized
in any way.


tlatorre-uchicago:
Can you resubmit and try doing this on an older version of XCode so that it doesn't add the animations tags?

tlatorre-uchicago:
I sent Mark Howe an email about this and he pointed me her: http://orca.physics.unc.edu/Orca_Help/Important_10.8_News.html.

I agree it would be nice to fix it in the GUI so that we don't have to update the settings for any new computers that we run on, but I don't want to merge this without Mark's approval.

Can you send him an email and explain exactly what the changes you made here are and see if he is OK with them?

PR #177, opened 2016-12-06
SNO+: Change the SNOPDetector view to a webview

marzece:


tlatorre-uchicago:
Can you resubmit without changing the old SNO nib file?

marzece:
Done

tlatorre-uchicago:
There is still a function in the SNO+ model that posts a notification of type viewTypeChanged which is now no longer registered in the SNO+ controller. Can you verify that this was only used for the detector state tab and if so, delete that from the SNO+ model?

marzece:
Yeah, you were talking about "ORSNOPModelViewTypeChanged" notification right? No thing else used that, so it's deleted now.

tlatorre-uchicago:
merged.

Issue #175, opened 2016-12-06
out of bounds array access in SNO+ controller


tlatorre-uchicago:
I was doing some tests on the teststand and was testing out the high thresholds button on the main SNO+ GUI and I got an out of bounds array access in displayThresholdsFromDB. Here is the debugger output:

```
2016-12-05 21:21:53.893 Orca[88170:303] Invalid color System, labelColor (warning given only once)
2016-12-05 21:21:54.010 Orca[88170:303] 10.4-style NSDateFormatter method called on a 10.0-style formatter, which doesn't work. Break on _NSDateFormatter_Log_New_Methods_On_Old_Formatters to debug. This message will only be logged once.
2016-12-05 21:21:54.215 Orca[88170:303] Could not find image named 'CameraPic'.
2016-12-05 21:22:07.093 Orca[88170:1150f] ERROR: unable to get the receiver data from the DB!
2016-12-05 21:22:51.940 Orca[88170:303] *** -[__NSArrayI objectAtIndex:]: index 0 beyond bounds for empty array
2016-12-05 21:22:51.947 Orca[88170:303] (
	0   CoreFoundation                      0x907f0471 __raiseError + 193
	1   libobjc.A.dylib                     0x9904b091 objc_exception_throw + 162
	2   CoreFoundation                      0x906c6fbc -[__NSArrayI objectAtIndex:] + 204
	3   Orca                                0x00931318 -[SNOPController displayThresholdsFromDB] + 13768
	4   Orca                                0x0092143e -[SNOPController SRVersionChanged:] + 382
	5   Foundation                          0x92be0992 __57-[NSNotificationCenter addObserver:selector:name:object:]_block_invoke + 49
	6   CoreFoundation                      0x907b95a4 __CFNOTIFICATIONCENTER_IS_CALLING_OUT_TO_AN_OBSERVER__ + 20
	7   CoreFoundation                      0x9069b04b _CFXNotificationPost + 3435
	8   Foundation                          0x92bcf47f -[NSNotificationCenter postNotificationName:object:userInfo:] + 92
	9   Foundation                          0x92beca3e -[NSNotificationCenter postNotificationName:object:] + 55
	10  Orca                                0x0093eca6 -[SNOPModel setStandardRunVersion:] + 166
	11  Orca                                0x00928f7a -[SNOPController runsLockChanged:] + 346
	12  Foundation                          0x92be0992 __57-[NSNotificationCenter addObserver:selector:name:object:]_block_invoke + 49
[...]

PR #181, opened 2016-12-06
Load high thresholds no matter what

jcaravaca:
Bug fix: high thresholds were only loaded if run ongoing. Simplify the load high threshold logic: instead of stopping and restarting the run, the action will set high threshold regardless the run status. This will change trigger settings of the ongoing run, which then will be thrown away by data quality. Nevertheless, this is an emergency action and if someone click this button is most likely because the ongoing run is misbehaving and won't be used in any case.

### DIFF
--- a/Source/Experiments/SNOP/SNOPController.m
+++ b/Source/Experiments/SNOP/SNOPController.m
@@ -360,21 +360,18 @@ snopGreenColor;
 {
     
     NSString* standardRun = [model standardRunType];
-    if(standardRun != nil && ![standardRun isEqualToString:@""] && [standardRunPopupMenu indexOfItemWithObjectValue:standardRun] == NSNotFound){
+    if([[model standardRunType] isEqualToString:@"HIGH THRESHOLDS"]) {
+        [model setStandardRunVersion:@"DEFAULT"];
+    }
+    else if(standardRun != nil && ![standardRun isEqualToString:@""] && [standardRunPopupMenu indexOfItemWithObjectValue:standardRun] == NSNotFound){
         NSLogColor([NSColor redColor],@"Standard Run \"%@\" does not exist in DB. \n",standardRun);
         return;
     }
     else{
         [standardRunPopupMenu selectItemWithObjectValue:standardRun];
-    }
-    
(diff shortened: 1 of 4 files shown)


jcaravaca:
This should fix #175. It's not totally clear why that was happening but I suspect that it was because the SR panel didn't find a "HIGH THRESHOLD" run in the DB. That logic is removed now and the High Threshold run is not treated as a usual SR anymore.

tlatorre-uchicago:
So, the "SET MTC HIGH THRESHOLDS" should now just set the high thresholds right?

Can you delete all the run control stuff from the loadHighThresholdRun function and maybe rename it since it now doesn't correspond to a run.

Also, why is there still logic in the SNOPController to handle a HIGH THRESHOLDS run?

If I remember correctly from our discussion at the DAQ workshop, a maintenance run is essentially a high thresholds run, so the normal course of action for an operator when the detector starts acting funny is to switch to a maintenance run which will quiet things down a bit, and then try and debug the problem.

Can you comment on the purpose of the button? Is it just to provide an alternate way of raising the thresholds in the case that the run control isn't working?

I just had a look and this is currently setting all the thresholds to zero. This is in fact a pretty dangerous button because if things start going haywire and someone presses this button, the detector will go silent even though there are potential analog problems which could fry the electronics.

jcaravaca:
This button was the 'No network Run' button: if the network is down and the detector goes crazy we wanted these safe settings hardcoded. This can also be used for debug as well.
The values were originally set to 4095 but I was told that 0 means triggers to max... is 4095 the max raw values then?

jcaravaca:
Tested at the teststand. As soon as I converge in the hardcoded values with @tlatorre-uchicago this should be ready to go.

tlatorre-uchicago:
The numbers are correct, but this button does nothing to set the detector in a safe state. If things are going haywire and there are lots of triggers firing there is a potential to fry the electronics. If you hit this button it will make the detector go silent so that it seems like things are safe, but in fact it does *nothing* to actually protect the analog hardware.

I'll send an email to the daq group to get feedback.

tlatorre-uchicago:
merged.

PR #183, opened 2016-12-06
Add DIAGNOSTIC run

jcaravaca:
Same as Maintenance but without changing trigger settings


tlatorre-uchicago:
Can you comment on whether this was tested on the teststand?

jcaravaca:
At the end it turned to be a bit more complicated: I need to make sure that all the GUIs are unlocked when in Diagnostic run. This implied some changed in the lock logic and changes in the controllers of ALL the SNO+ objects.
Tested OK at the teststand.

jcaravaca:
I noticed that the Hardware Wizard will need to be changed as well but that'd mess up with core ORCA. I contacted Mark and he will implement a variable mask that can be defined by the individual experiments so that we can have Maintenance and Diagnostic runs to unlock the GUIs, since that's not in ORCA by design. I'll probably have to revisit the whole logic so hold the merge until I add then.

jcaravaca:
Mark updated ORCA to allow unlock the Wizard using another mask. @tlatorre-uchicago could you resync with svn so that I can implement this?

BenLand100:
@tlatorre-uchicago fyi I rebased from svn and pushed master

jcaravaca:
Tested at the test stand. Treat runs with the 'DIAGNOSTIC' bit up as Maintenance but without changing the detector settings at the start of a run. This PR should be ready to go.

tlatorre-uchicago:
I will test this out today during my shift and merge. Thanks!

tlatorre-uchicago:
merged.

Issue #188, opened 2016-12-08
unable to turn HV on on teststand


tlatorre-uchicago:
Here are the steps that happened (unknown if this is reproducible):

1. powered on crate
2. Hit crate reset
3. Crate reset was successful (in daq logs) but ORCA timed out

At this point ORCA reconnected to the XL3 but the HV tab doesn't show anything and all the HV buttons are disabled. I verified that the HV thread is indeed still running.

![screen shot 2016-12-07 at 9 07 52 pm](https://cloud.githubusercontent.com/assets/5834059/20994982/eece1936-bcc1-11e6-9f37-8036ffc025cd.png)

tlatorre-uchicago:
Fixed in #190.

PR #190, opened 2016-12-08
SNO+: Fix user-lockout of HV control

BenLand100:
The methods to read back voltage/current and switch status can fail and return without throwing an exception in certain cases. If reading switch status failed in this way but then reading back voltage was successful, the hv control logic would end up in a state where the monitoring thread was running but the user was unable to control the high voltage. This ensures that the switch and voltage are both successfully read back before starting the high voltage thread, preventing the above from occurring.

In #188 this was realized by the xl3 disconnecting and reconnecting quickly. Between checking that the XL3 was connected and reading the switch / voltage status the XL3 disconnected. When the HV monitoring thread was launched the XL3 had connected again and did read back the voltage eventually, but the hv monitoring thread does not query the switch status again so the GUI correctly assumed it did not know the XL3 state and refused user control.


tlatorre-uchicago:
merged.

PR #197, opened 2016-12-11
HV Improvements

BenLand100:
Introduces a loopCounter in the hv thread (where loops are ~1s each) such that the heartbeat is sent only on loopCounter multiples of 10. Also modifies HV alarms to be updated on the initial loop (and 2^32 loops) or whenever the state of the alarm changes which should significantly reduce db load.

### DIFF
--- a/Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.h
+++ b/Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.h
@@ -108,12 +108,15 @@ enum {
     BOOL hvBSwitch;
     BOOL hvARamping;
     BOOL hvBRamping;
+    BOOL hvAQueryWaiting;
+    BOOL hvBQueryWaiting;
     BOOL hvAFromDB;
     BOOL hvBFromDB;
     BOOL hvEverUpdated;
     BOOL hvSwitchEverUpdated;
     BOOL hvANeedsUserIntervention;
     BOOL hvBNeedsUserIntervention;
+    
     
     NSString* triggerStatus;
     BOOL _isTriggerON;
@@ -253,6 +256,8 @@ enum {
 @property float vhighalarm_b_vmax;
 @property float ihighalarm_b_imax;
 
+@property BOOL hvAQueryWaiting;
+@property BOOL hvBQueryWaiting;
 @property BOOL hvAFromDB;
 @property BOOL hvBFromDB;
 @property BOOL hvEverUpdated;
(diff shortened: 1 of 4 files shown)


BenLand100:
@tlatorre-uchicago I won't have time to test this today but I can test on the teststand tomorrow. I don't expect any surprises but it would be good to check that the heartbeats are still being issued and that alarms still get issued since I modified all that code.

BenLand100:
@tlatorre-uchicago I included three distinct things in this PR since they all needed testing together - would suggest looking at the diffs for each commit separately though. The first reduces the heartbeat update and changes level alarms to only write to the DB when the state changes. The second implements what you requested re: NSAutoreleasePool inside the HV loops. Finally I updated the db parameter readback with your comments (inform user if no postgres object, ask user to use hardcoded values if can't access db) however I expect this will be hard to follow since it was tricky to implement. I've updated the test and detector db's with a schema with limits:
```
CREATE TABLE hvparams (
    crate           smallint NOT NULL,
    supply          AorB NOT NULL DEFAULT 'A',
    nominal         real NOT NULL CHECK (nominal >= 0),         --volts - nominal voltage
    hvramp_up       real NOT NULL CHECK (hvramp_up > 0),        --volts / second - ramp up speed
    hvramp_down     real NOT NULL CHECK (hvramp_down > 0),      --volts / second - ramp down speed
    vsetalarm_vtol  real NOT NULL CHECK (vsetalarm_vtol >= 0),  --volts - tolerance of voltage setpoint alarm
    ilowalarm_vmin  real NOT NULL CHECK (ilowalarm_vmin >= 0),  --volts - minimum voltage required to check for current
    ilowalarm_imin  real NOT NULL CHECK (ilowalarm_imin >= 0),  --microamps - current below this when voltage is above ilowalarm_vmin will trigger a current dropout alarm
    ihighalarm_imax real NOT NULL CHECK (ihighalarm_imax >= 0), --microamps - current above this will trigger an overcurrent alarm
    vhighalarm_vmax real NOT NULL CHECK (vhighalarm_vmax >= 0), --volts - voltage above this will trigger an overvoltage alarm
    PRIMARY KEY (crate, supply)
);
```
The current limits are set to reasonable values for a typical crate (65mA max, at least 1mA above 100V) while other db parameters have been updated to reflect what was last tested on the detector. 

Also fixed minor issues with button and message text being swapped for some popups, missing newlines in NSLog, and such.

Tested crate 19 hv alarms on the teststand along with everything else, seems to be working as expected. Ready for merge after testing on detector.

* Compiled on ancient xcode
* Checked for memory leaks
* Didn't modify gui files
* Tested on teststand 12/12

tlatorre-uchicago:
Awesome! I will test this out today during my shift and merge.

BenLand100:
@tlatorre-uchicago I added an NSLog at the start of the _hvXl3 thread - it's one line printout and a little long, but probably fine. Also added the check to keep from looping through reading the switches out.

I checked quickly on the teststand so should be no surprises.

tlatorre-uchicago:
merged.

PR #198, opened 2016-12-13
Attempt to fix the SNO+ GUI freeze-out by sending the GUI changes to the main thread

jcaravaca:


jcaravaca:
From PR #173: I divided them in this one and another one containing only the unrelated bug fixes.

Source/Experiments/SNOP/SNOPController.m
```diff
@@ -501,17 +501,17 @@ - (void) viewTypeChanged:(NSNotification*)aNote
...
 - (void) dbDebugDBIPChanged:(NSNotification*)aNote
-{
+{ dispatch_async(dispatch_get_main_queue(), ^{
     [debugDBIPAddressPU setStringValue:[model debugDBIPAddress]];
-}
+}); }
 
 - (void) hvStatusChanged:(NSNotification*)aNote
-{
+{ dispatch_async(dispatch_get_main_queue(), ^{
```
tlatorre-uchicago: I went through the XL3 code and this notification is only posted on the main thread.

jcaravaca: It seems that's true as well for the runControl and the ORDetector base class... what it's odd to me is that when these lines were in place during last running, the problem seemed to disappear. Also, the XL3 code had some freeze-out problem (maybe not related though) that was fixed by dispatching the changes to the main thread. @BenLand100 can confirm this.

PR #199, opened 2016-12-13
SNO+: Bug fixes.

jcaravaca:
Prevent crashes when SR type and version not correctly set. Stop running ORCA script when stopping runs.

PR by Javier Caravaca

### DIFF
--- a/Source/Experiments/SNOP/SNOPController.m
+++ b/Source/Experiments/SNOP/SNOPController.m
@@ -71,6 +71,14 @@ snopGreenColor;
     
     [super dealloc];
 }
+
+- (IBAction) orcaDBTestAction:(id)sender {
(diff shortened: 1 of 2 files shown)


tlatorre-uchicago:
@jcaravaca, can you remove the purely stylistic changes from this?

jcaravaca:
@tlatorre-uchicago I rearranged few lines as I went to get the code more tidy and readable. It doesn't affect performance, is there any problem with keeping them?

tlatorre-uchicago:
The two I have issue with is the detector tile one because that is going to cause a merge conflict and the debugdbtestaction since it no longer fits in 80 characters.

tlatorre-uchicago:
merged.

PR #201, opened 2016-12-14
Triggers on warning

tlatorre-uchicago:
This pull request adds a dialog box which pops up when the user enables triggers if the crate is not at HV asking them if they are sure they want to turn triggers on.


tlatorre-uchicago:
merged.

PR #203, opened 2016-12-14
SNO+: update XL3 model to wait until crate has been initialized before checking the HV

tlatorre-uchicago:
The HV registers are undefined just after the crate has been turned on but
before the crate has been reset. So, the HV thread now waits until the crate
has been reset before continuing. This commit also updates the crate reset
button so that it is enabled only when either the crate has not been reset or
if the HV is off.

@BenLand100 can you just take a quick look at my change to the _hvInit method and make sure that it makes sense?

### DIFF
--- a/Source/Objects/Custom Hardware/SNO+/SNOCrate/ORSNOCrateController.m
+++ b/Source/Objects/Custom Hardware/SNO+/SNOCrate/ORSNOCrateController.m
@@ -69,26 +69,28 @@
 
     ORXL3Model *xl3 = [model adapter];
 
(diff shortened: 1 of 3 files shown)


tlatorre-uchicago:
Tested at the teststand on Wednesday, December 14, 2016.

- Verified that the crate reset button is enabled just after a power cycle
- Verified that the HV thread doesn't start until after a crate reset
- Verified that the crate reset button is greyed out after HV is turned on
- Verified that the crate reset button is enabled and the HV thread both don't start even after a power cycle with ORCA running

tlatorre-uchicago:
merged.

PR #207, opened 2016-12-16
SNO+: add a ping crates button to the SNO+ Control Panel

tlatorre-uchicago:
This button enables pedestals for one crate at a time and sends a single
pedestal pulse. This is useful to make sure that you are getting NHIT triggers
from all the crates.

This has been tested on the teststand on Friday, December 16, 2016. I verified that I could successfully ping crate 19, although since this function is supposed to go crate by crate it definitely needs testing on the full detector to make sure that it is doing what we expect it to do.

### DIFF
--- a/Source/Experiments/SNOP/SNOP.xib
+++ b/Source/Experiments/SNOP/SNOP.xib
@@ -1,10 +1,10 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="6254" systemVersion="13F1077" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none">
+<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="6751" systemVersion="13F1603" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none">
     <dependencies>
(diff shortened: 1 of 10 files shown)


jcaravaca:
Do we want this capability while not running in Maintenance? I would assume so but if not we should lock this button when running.

Certainly we don't want to change pedestal masks when taking ECAs so could you include a check for the ECA bits not being set in the run type word?

tlatorre-uchicago:
I'm not 100% sure when we want to allow this button. Josh originally suggested potentially running it by default at the start of every run, so I think it's OK to allow it during a physics run.

I will add a function to disable the button during ECA runs.

nbarros:
How hard would it be to do it as a sort of pre-run procedure? You press start run, the pings are sent and then the run starts. 
I imagine that it would make the "restart" tricky, but would hard stop/start and resync be an issue?

tlatorre-uchicago:
That would be fairly easy, except that I would need to update the function to reset the original pedestal mask after it finishes (otherwise it would be really annoying if you were pedestaling certain crates and then when a new run started, the pedestal masks all went to 0).

In fact, there is a period during hard run starts that we specifically allocate for changing hardware (it's before the first SOFT_GT which marks the start of good data). I'll have to check and see how easy it would be to do during this period, but that would be perfect.

tlatorre-uchicago:
merged. I will open an issue about whether we want to do this at the start of every run.

PR #209, opened 2016-12-17
SNO+: delete code in the SNO+ model which uploads the detector conigfuration, mtc configuration, and run documents to CouchDB

tlatorre-uchicago:
The detector state is now saved to a PostgreSQL database and is handled outside
of ORCA.

Tested on the detector on Friday, December 16.


tlatorre-uchicago:
merged.

PR #210, opened 2016-12-17
SNO+: fix bug which caused multiple dogs to appear in status bar

tlatorre-uchicago:
The timer which is used to animate the dog icon holds a reference to the RunStatusIcon class so it is not actually freed when the SNO+ controller is deallocated.

This commit fixes the issue by making sure to only create one timer (this was causing the dogs to run at different speeds I think) and then making sure to stop the animation before releasing the object.


tlatorre-uchicago:
merged.

Issue #106, opened 2016-05-10
Have to stop pulser when changing from PEDs to PGT


icoulter:
To use pulser for PGT instead of PEDs you have to change pulser feed, stop the pulser then load and fire. Not the case for the reverse when changing pulser feed and clicking load+fire is all thats required.

tlatorre-uchicago:
Fixed in #212.

PR #212, opened 2016-12-17
SNO+: fix the MTC model so that it disables pedestals when pressing the "Load & Fire" button if the pulser feed is set to PGT

tlatorre-uchicago:
Fixes issue #106.


tlatorre-uchicago:
merged.

PR #217, opened 2016-12-17
SNO+: update HV check in reset crate function

tlatorre-uchicago:
This pull request updates the check in the reset crate function to make sure that the HV thread has updated the switch values before checking if the switches are on.

I noticed this on the teststand when I power cycled the crate with the HV switch still on. When the XL3 came back the HV thread didn't start (because the crate hadn't been reset), but the switch variables in ORCA were still set to show that the HV was on so I had no way to reset the crate.

Tested on the teststand on Saturday, December 17, 2016.


tlatorre-uchicago:
merged.

PR #225, opened 2016-12-22
SNO+: Fix supply A overcurrent alarm not clearing.

BenLand100:
Found by @fbdescamps and crew, a typo set the last state of supply B instead of supply A so the alarm was triggered but when the state cleared internally he DB was never updated. This fixes the typo.

Simple change, so should be safe to merge, but hasn't actually been tested.


BenLand100:
merged.

Issue #230, opened 2017-01-05
Change colors of trigger thresholds in Standard Run tab for colorblind users


jackdunger:
At @ecaden's request. They are currently red/green, red/blue like the rest of the GUI could be safer.

tlatorre-uchicago:
I would vote for bold and grey for enable/disabled.

jcaravaca:
We've used red/blue throughout ORCA and it'd be nice to be consistent, but I can change it if some other people chime on this

tlatorre-uchicago:
Fixed in #234.

Issue #204, opened 2016-12-14
Coloured run bar displays 'Running in Maintenance' when in diagnostic


EdLeming:
In the SNOPController, the coloured run bar display does not agree with the run type field.

tlatorre-uchicago:
Fixed in #234.

PR #234, opened 2017-01-08
SNO+: Additions to SNO+ GUI.

### LINKED ISSUE #227: SNO+ control panel: Run Type word protection
The operator can change any radio-button state in the SNO+ main GUI run-type word. The only ones that an operator should be allowed to change is the bottom set of 'detector state' .  The other bits should be set depending on which run is selected.

Also, the spare bits should be greyed out and not selectable.

jcaravaca:
Fixes #227, #230 and #204 
Also ask the user to refresh the SR if not set and lock RESYNC button if run is not ongoing.


jcaravaca:
Tested at TestStand and ready to go. It'd be nice to have this merged ASAP to unlock the SNOP GUI by beginning of next week.

tlatorre-uchicago:
merged.

PR #235, opened 2017-01-09
SNO+: Add Run Type Word Headers to be shared with the RAT code

jcaravaca:


tlatorre-uchicago:
I prefer the old style of defining the run types using the mask so that you can check the run type with a simple and statement:

```
if (run_type & DIAGNOSTIC_RUN)
```

This is how the run type is defined in XSNOED and all the daq code.

jcaravaca:
@tlatorre-uchicago I can change that, but I'd prefer to keep the enum rather than the define statements that there were in the old version

tlatorre-uchicago:
Sure, that sounds good.

jcaravaca:
Just let me know if you are happy with the changes but **do not merge yet**. I have to wait for approval of the RAT CIC as well.

jcaravaca:
CIC has merged this already into RAT so it should be ready to go now.

tlatorre-uchicago:
merged.

Issue #222, opened 2016-12-20
Nominal Voltage shown for all crates in HV Master in Control Panel


ecaden:
Instead of having the nominal voltage shown only when the rack is on, each crate's nominal voltage should always be shown on the HV master tab. Maybe some color change is appropriate if the rack is then on.

BenLand100:
@jcaravaca on the HV master tab would it be difficult to add logic to display a nominal voltage of 0 as 'OFF' or 'D/C' or 'N/A' and @ecaden would that satisfactorily solve this issue? The reason it's showing up as 0 now is the code to check the DB happens as the XL3's connect to avoid flooding queries right when orca starts - so crates that are off never query the DB.

ecaden:
The nominal voltage of each rack shouldn't change, I'm not sure why this is coming from the database? I would think it should be hard coded, also as the limiting voltage that each crate could ramp to. 

--
Erica Caden

BenLand100:
I have heard elsewhere that the nominal voltages are not set in stone - mostly because PMTs have been swapped out and there is a variation in base resistance. Since it's something that could potentially change having it compiled into ORCA is less than ideal because there's not really a substantial difference in having it hardcoded vs read from the database except that the former requires a recompile to change it. Maybe this is something to bring up on a DAQ call.

tlatorre-uchicago:
Fixed in @243.

Issue #165, opened 2016-12-01
HV thread doesn't stop when experiment is closed


tlatorre-uchicago:
When closing the experiment in ORCA to switch to a new .Orca file, sometimes the ORXL3Model segfaults in readHVStatus. Here is the debugger output:

```
2016-12-01 16:54:08.301 Orca[49818:303] Invalid color System, labelColor (warning given only once)
2016-12-01 16:54:08.589 Orca[49818:303] Could not find image named 'CameraPic'.
2016-12-01 16:54:33.463 Orca[49818:303] *** Assertion failure in -[NSTextFieldCell _objectValue:forString:errorDescription:], /SourceCache/AppKit/AppKit-1265.21/AppKit.subproj/NSCell.m:1637
2016-12-01 16:54:33.466 Orca[49818:303] *** Assertion failure in -[NSTextFieldCell _objectValue:forString:errorDescription:], /SourceCache/AppKit/AppKit-1265.21/AppKit.subproj/NSCell.m:1637
2016-12-01 16:54:33.522 Orca[49818:303] *** Assertion failure in -[NSTextFieldCell _objectValue:forString:errorDescription:], /SourceCache/AppKit/AppKit-1265.21/AppKit.subproj/NSCell.m:1637
2016-12-01 16:54:34.293 Orca[49818:1c61f] -[__NSCFString crateNumber]: unrecognized selector sent to instance 0x2a00f0b0
2016-12-01 16:54:34.294 Orca[49818:1c61f] An uncaught exception was raised
2016-12-01 16:54:34.294 Orca[49818:1c61f] -[__NSCFString crateNumber]: unrecognized selector sent to instance 0x2a00f0b0

2016-12-01 16:54:34.304 Orca[49818:1c61f] (
	0   CoreFoundation                      0x907f0471 __raiseError + 193
	1   libobjc.A.dylib                     0x9904b091 objc_exception_throw + 162
	2   CoreFoundation                      0x907f4cb3 -[NSObject(NSObject) doesNotRecognizeSelector:] + 275
	3   CoreFoundation                      0x90740522 ___forwarding___ + 1010
	4   CoreFoundation                      0x9074010e _CF_forwarding_prep_0 + 14
	5   Orca                                0x00889440 -[ORXL3Model crateNumber] + 48
	6   Orca                                0x0089b2ef -[ORXL3Model readHVStatus] + 239
	7   Orca                                0x008a6f04 -[ORXL3Model(private) _hvXl3] + 4948
[...]

PR #243, opened 2017-01-16
SNO+: Load HV params eariler, constrain HV mon thread lifetime.

BenLand100:
ORXL3Model now listens for document opened and closed notifications
and loads hv parameters asap after the document is loaded to give
time for the postgres object to become available. Additionally the
hv monitoring thread now exits when the document is closed to
avoid crashes at orca shutdown.

Pending testing on teststand, etc.

### DIFF
--- a/Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.h
+++ b/Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.h
@@ -116,6 +116,7 @@ enum {
     BOOL hvSwitchEverUpdated;
     BOOL hvANeedsUserIntervention;
     BOOL hvBNeedsUserIntervention;
+    BOOL isLoaded; //Whether the experiment is open or not (false to start, false at end)
     
     
     NSString* triggerStatus;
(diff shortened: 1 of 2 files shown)


BenLand100:
Fixes #222 and #165

BenLand100:
Tested on teststand 1/21 - these changes shouldn't require additional detector testing as hvparams are now loaded for all crates after the experiment opens and the changes to the hvthread itself are minor and validated with crate 19.

BenLand100:
@tlatorre-uchicago assuming the last comment seems reasonable this is good to merge.

tlatorre-uchicago:
I tried this out on the teststand and noticed that if the database is unavailable and you click Try Again you can't do anything in the GUI for ten seconds. It looks like the sleep function is being run on the main thread.

BenLand100:
That's not strictly speaking due to this PR - I believe Phil made a change that makes database callbacks run on the main thread and there is a sleep in there. Probably the best option is to just remove that 10s sleep as it's not doing anything critical. I'll push momentarily.

tlatorre-uchicago:
merged.

Issue #228, opened 2017-01-04
Does not correctly set XL3 slot mask when setting XL3 mode.


bonventre:
It appears like the XL3 slot mask is set to slots present regardless of the slot mask set in the XL3 window. It is also not clear from the gui that this is only set when the mode is set to normal.

tlatorre-uchicago:
This was an intentional change, but maybe we need to revisit it. The problem was that the set XL3 mode button is very far away from the slot mask and most people did not really understand that when you set the XL3 to normal mode it only reads out those channels masked in above which were often different from the cards present because those same check marks are used to set the pedestal mask.

Not sure what the best solution is here.

tlatorre-uchicago:
Fixed in #254.

PR #254, opened 2017-01-19
SNO+: update load hardware button to put XL3 in NORMAL mode

tlatorre-uchicago:
Also, allow the user to change the readout mask when setting the mode from the
Ops tab in the XL3 window.

Fixes issue #228.


tlatorre-uchicago:
Tested on the teststand on Wednesday, January 18.

tlatorre-uchicago:
merged.

PR #255, opened 2017-01-19
SNO+: Add ECA run.

jcaravaca:
Features:
- Get rid of the ORCA Script dependency that was causing troubles.
- Launch ECA runs in a different thread.
- Stop ECA run smothly when restart/resync/stop run.
- Rollover to previous run if any.


Source/Experiments/SNOP/ECARun.m
```diff
@@ -0,0 +1,452 @@
...
+- (void) doECAs
+{
+    @autoreleasepool {
+        
+        //Get MTC model
+        NSArray* objs = [[(ORAppDelegate*)[NSApp delegate] document] collectObjectsOfClass:NSClassFromString(@"ORMTCModel")];
+        if ([objs count]) {
+            anMTCModel = [objs objectAtIndex:0];
+        } else {
+            NSLogColor([NSColor redColor], @"ECARun: couldn't find MTC model. \n");
```
tlatorre-uchicago: Need to bomb out here?

Source/Experiments/SNOP/ECARun.m
```diff
@@ -0,0 +1,452 @@
...
+        }
+        
+        //Get XL3 models
+        anXL3Model = [[(ORAppDelegate*)[NSApp delegate] document] collectObjectsOfClass:NSClassFromString(@"ORXL3Model")];
+        if (![anXL3Model count]) {
+            NSLogColor([NSColor redColor], @"ECARun: couldn't find XL3 model. \n");
```
tlatorre-uchicago: here too

Source/Experiments/SNOP/ECARun.m
```diff
@@ -0,0 +1,452 @@
...
+        }
+        
+        //Get FEC models
+        aFECModel = [[(ORAppDelegate*)[NSApp delegate] document] collectObjectsOfClass:NSClassFromString(@"ORFec32Model")];
+        if (![aFECModel count]) {
+            NSLogColor([NSColor redColor], @"ECARun: couldn't find XL3 model. \n");
```
tlatorre-uchicago: here too

tlatorre-uchicago: message should be FECs, not XL3

Source/Experiments/SNOP/ECARun.m
```diff
@@ -0,0 +1,452 @@
...
+        }
+        
+        //Check pulser rate is not zero
+        if([ECA_rate isEqual:0]){
+            NSLogColor([NSColor redColor], @" Set Pulser Rate to a non-zero value \n");
+            goto stop;
+        }
+        
+        //Wait for run to start
+        sleep(1.);//wait so that
```
tlatorre-uchicago: wait so that what?

Source/Experiments/SNOP/ECARun.m
```diff
@@ -0,0 +1,452 @@
...
+        
+        NSLog(@"************************* \n");
+        NSLog(@"Finishing ECA Run \n");
+        NSLog(@"************************* \n");
+        
+        //Unset pedestal masks
+        [aSNOPModel zeroPedestalMasks];
+        
+        //Rollover to previous run, if any. Otherwise stop run.
+        if(rollover){
```
tlatorre-uchicago: technically not a rollover, maybe call it start_new_run

Source/Experiments/SNOP/ECARun.m
```diff
@@ -0,0 +1,452 @@
...
+    
+    //EPED headers
+    [aSNOPModel updateEPEDStructWithCoarseDelay: 255 - floor(coarse_delay/10.0) fineDelay: fine_delay chargePulseAmp: 0x0	pedestalWidth: pedestal_width calType: 10 + (ECA_pattern+1)];
+    
+    //Inject pedestals
+    int eca_pattern_num_steps = getECAPatternNSteps(ECA_pattern);
+    NSMutableArray* pedestal_mask = getECAPattern(ECA_pattern);
+    if ([pedestal_mask count] == 0 || pedestal_mask == nil) {
+        NSLogColor([NSColor redColor], @"ECA: Error loading pedestal mask for pattern: %d \n", ECA_pattern);
+        return false;
```
tlatorre-uchicago: don't release alloc'ed arrays above.

Source/Experiments/SNOP/ECARun.m
```diff
@@ -0,0 +1,452 @@
...
+            //Fire pedestals at the pulser rate during a time=sub_run_time
+            [anMTCModel continueMTCPedestalsFixedRate];
+            usleep([self ECA_subruntime] * 1e6);
+            
+            //End of sub-run
+            [anMTCModel stopMTCPedestalsFixedRate];
+            
+            //Check if user has cancel ECAs and quit in that case
+            if([ECAThread isCancelled]) {
+                return false;
```
tlatorre-uchicago: need to release memory

Source/Experiments/SNOP/ECARun.m
```diff
@@ -0,0 +1,452 @@
...
+            }
+            
+        }
+
```
tlatorre-uchicago: here too.

tlatorre-uchicago: Also, if any of the above methods can raise an exception, need to catch it here and make sure the memory is released.

Source/Experiments/SNOP/ECARun.m
```diff
@@ -0,0 +1,452 @@
...
+        //NSLog(@"PEDESTAL MASK: %d, %d, %x \n",crate_number,card_number,mask);
+        [fec setPedEnabledMask:mask];
+    }
+
+    //Load pedestal mask
+    for (ORXL3Model *xl3 in anXL3Model) {
+        //NSLog(@"Set pedestal for XL3 %d \n", [xl3 crateNumber]);
+        dispatch_async(dispatch_get_main_queue(), ^{
+            [xl3 setPedestalInParallel];
+        });
```
tlatorre-uchicago: need to find a way to make sure that the pedestals have actually finished being set before returning.

Source/Experiments/SNOP/ECARun.m
```diff
@@ -0,0 +1,452 @@
...
+    [aSNOPModel updateEPEDStructWithCoarseDelay: 255 - floor(coarse_delay/10.0) fineDelay: fine_delay chargePulseAmp: 0x0	pedestalWidth: pedestal_width calType: 10 + (ECA_pattern+1)];
+
+    //Inject pedestals
+    int eca_pattern_num_steps = getECAPatternNSteps(ECA_pattern);
+    NSMutableArray* pedestal_mask = getECAPattern(ECA_pattern);
+    if ([pedestal_mask count] == 0 || pedestal_mask == nil) {
+        NSLogColor([NSColor redColor], @"ECA: Error loading pedestal mask for pattern: %d \n", ECA_pattern);
+        return false;
+    }
+    [anMTCModel fireMTCPedestalsFixedRate];
```
tlatorre-uchicago: I don't think this is thread safe.

Source/Experiments/SNOP/ECARun.m
```diff
@@ -0,0 +1,452 @@
...
+        [self changePedestalMask:pedestal_mask[step]];
+        
+        //Inject subrun (we don't do this anymore, but leave it here until
+        //tested at the detector)
+        //[aRunModel prepareForNewSubRun];
+        [aSNOPModel updateEPEDStructWithStepNumber:step];
+        //[aRunModel startNewSubRun];
+
+        //Ship EPED Headers
+        [aSNOPModel shipEPEDRecord];
```
tlatorre-uchicago: not thread safe.

Source/Experiments/SNOP/ECARun.m
```diff
@@ -0,0 +1,452 @@
...
+        
+        //Fire pedestals at the pulser rate during a time=sub_run_time
+        [anMTCModel continueMTCPedestalsFixedRate];
```
tlatorre-uchicago: not thread safe.

Source/Experiments/SNOP/ECARun.m
```diff
@@ -0,0 +1,452 @@
...
+        usleep([self ECA_subruntime] * 1e6);
+        
+        //End of sub-run
+        [anMTCModel stopMTCPedestalsFixedRate];
```
tlatorre-uchicago: not thread safe.

Source/Experiments/SNOP/ECARun.m
```diff
@@ -0,0 +1,452 @@
...
+{
+
+    //Set pedestal mask in FEC: this does not load it yet
+    for (ORFec32Model *fec in aFECModel) {
+        int crate_number = [fec crateNumber];
+        if(crate_number == 19) crate_number = 0; //hack for the teststand
+        int card_number = [fec stationNumber];
+        int mask = [[[pedestal_mask objectAtIndex:crate_number] objectAtIndex:card_number] intValue];
+        //NSLog(@"PEDESTAL MASK: %d, %d, %x \n",crate_number,card_number,mask);
+        [fec setPedEnabledMask:mask];
```
tlatorre-uchicago: not thread safe.

Source/Experiments/SNOP/ECARun.m
jcaravaca: Don't get this. Where is the problem?

Source/Experiments/SNOP/ECARun.m
jcaravaca: Should this be accessed by the main thread rather?

Source/Experiments/SNOP/ECARun.m
tlatorre-uchicago: If you can't find the MTC model then the function continues executing and will segfault.

Source/Experiments/SNOP/ECARun.m
tlatorre-uchicago: Yes. That or you need to create thread safe functions in the MTC model.

jcaravaca: We shouldn't call this method (and the ones below) from the main thread since we lost the aim of this implementation. About making the MTC model thread-safe, what is the exact concern? I thing the risk of accessing the same function from two different threads is unlikely when taking ECAs.

Source/Experiments/SNOP/ECARun.m
BenLand100: http://www.enekoalonso.com/articles/running-multiple-asynchronous-simultaneously-and-waiting-for-completion
see parts related to dispatch_group_wait

BenLand100: (unfortunately that code is in swift but the functions all have the same names in objective-c)

Source/Experiments/SNOP/ECARun.m
tlatorre-uchicago: It has to be called from the main thread until the method is made thread safe. It may still be faster than the orca script which presumably has to execute everything on the main thread whereas here you can only execute those methods which are necessary to run on the main thread. Plus, in the future if we make the MTC methods thread safe (which will require first making the RedisClient object thread safe), we can remove them.

jcaravaca: OK, we'll live with that so far and I'll give it a try on site to see whether it solves the issue of the orca scripts

jcaravaca:
Tested OK at the TS 1/26:
- [x] Started/stopped several ECA runs while stopped/running. ECAs finish smoothly if stopped by the user and rollover happens if previous run ongoing.
- [x] Check with XSnoed we were applying the proper ped mask and getting the correct pulser rate.
- [x] Compiles in XCode5.

From my side this should be ready to go. I'd like to test it onsite first, so once we're all happy I'll deploy this branch in the detector to check whether everything works as expected.

Source/Experiments/SNOP/ECARun.m
```diff
@@ -0,0 +1,565 @@
...
+{
+    return [ECAThread isFinished];
+}
+
+- (void) setPulserRate:(NSNotification*)aNote
+{
+    
+    /* Ensure we are in the main Thread since this method
+     is called from a Notification */
+    dispatch_async(dispatch_get_main_queue(), ^{
```
tlatorre-uchicago: If this is a notification posted from the run model it is guaranteed to be run on the main thread. The asynchronous dispatch here won't guarantee that the pulser rate is actually changed by the time the run starts.

Source/Experiments/SNOP/ECARun.m
```diff
@@ -0,0 +1,565 @@
...
+    });
+
+}
+
+- (void) launchECAThread:(NSNotification*)aNote
+{
+    
+    /* Ensure we are in the main Thread since this method
+     is called from a Notification */
+    dispatch_async(dispatch_get_main_queue(), ^{
```
tlatorre-uchicago: same here.

Source/Experiments/SNOP/ECARun.m
```diff
@@ -0,0 +1,565 @@
...
+            
+            //Launch ECA thread
+            [ECAThread release];
+            ECAThread = [[NSThread alloc] initWithTarget:self selector:@selector(doECAs) object:nil];
+            [ECAThread start];
+            [[NSNotificationCenter defaultCenter] postNotificationName:ORECARunStartedNotification object:self];
+            
+        }
+        else{
+            //Nothing
```
tlatorre-uchicago: superfluous else statement?

Source/Experiments/SNOP/ECARun.m
```diff
@@ -0,0 +1,565 @@
...
+    @autoreleasepool {
+        
+        /* Get models */
+        //SNO+ model
+        NSArray* objs = [[(ORAppDelegate*)[NSApp delegate] document] collectObjectsOfClass:NSClassFromString(@"SNOPModel")];
+        if ([objs count]) {
+            aSNOPModel = [objs objectAtIndex:0];
+        } else {
+            NSLogColor([NSColor redColor], @"ECARun: couldn't find SNO+ model. \n");
+            goto stop;
```
tlatorre-uchicago: will segfault at stop since aSNOPModel is not defined.

Source/Experiments/SNOP/ECARun.m
```diff
@@ -0,0 +1,565 @@
...
+                }
+                
+            }
+            
+        } // end ECA steps
+    }
+    @catch(...){
+        //Don't leak memory
+        [tslope_delays_coarse release];
+        [tslope_delays_fine release];
```
tlatorre-uchicago: Do you need to release pedestal_mask here too?

tlatorre-uchicago: It might be easier to not alloc things explicitly. That way the autorelease pool will take care of releasing all the memory.

Source/Experiments/SNOP/ECARun.m
```diff
@@ -0,0 +1,565 @@
...
+            NSLogColor([NSColor redColor], @" Enable EXT PED bit in the MTC trigger mask. Stopping ECAs... \n");
+            goto stop;
+        }
+        
+        //Hold the current run until the ECAs are done
+        NSDictionary *userInfo = [NSDictionary dictionaryWithObjectsAndKeys:
+                                  @"waiting for ECA run to finish", @"Reason",
+                                  nil];
+        dispatch_async(dispatch_get_main_queue(), ^{
+            [[NSNotificationCenter defaultCenter] postNotificationName:ORAddRunStateChangeWait object: self userInfo: userInfo];
```
tlatorre-uchicago: This should be posted after the run is about to stop, not at the beginning.

Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
```diff
@@ -2690,7 +2696,9 @@ - (void) setPedestalInParallel
         return;
     }
 
-    [NSThread detachNewThreadSelector:@selector(_setPedestalInParallelWorker) toTarget:self withObject:nil];
+    NSThread *xl3Thread = [[NSThread alloc] initWithTarget:self selector:@selector(_setPedestalInParallelWorker) object:nil];
```
tlatorre-uchicago: Why did you change this line?

Source/Experiments/SNOP/ECARun.m
jcaravaca: this is actually not needed since it's already defined above

Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
jcaravaca: Didn't revert back the change it after some experiments. Should be unchanged now

jcaravaca:
@tlatorre-uchicago I've included your comments

Source/Experiments/SNOP/ECARun.m
```diff
@@ -0,0 +1,541 @@
...
+    }
+    else{
+        
+        //Get SNO+ model
+        NSArray* objs = [[(ORAppDelegate*)[NSApp delegate] document] collectObjectsOfClass:NSClassFromString(@"SNOPModel")];
+        if ([objs count]) {
+            aSNOPModel = [objs objectAtIndex:0];
+        } else {
+            NSLogColor([NSColor redColor], @"ECARun: couldn't find SNO+ model. \n");
+            return;
```
tlatorre-uchicago: There is no indication to the caller here that this failed and the rest of the functions will still segfault.

Source/Experiments/SNOP/ECARun.m
```diff
@@ -0,0 +1,541 @@
...
+    if(start_eca_run){
+        
+        /* If we starting an ECA run we have to set the pulser
+         rate enter by the operator, not the standard run one*/
+        //MTC model
+        NSArray *objs = [[(ORAppDelegate*)[NSApp delegate] document] collectObjectsOfClass:NSClassFromString(@"ORMTCModel")];
+        if ([objs count]) {
+            anMTCModel = [objs objectAtIndex:0];
+            [anMTCModel setDbFloat:[[self ECA_rate] floatValue] forIndex:kPulserPeriod]; //Set pulser rate in GUI
+            [anMTCModel setThePulserRate:[[self ECA_rate] floatValue]]; //UNCOMMENT FOR COMMISSION
```
tlatorre-uchicago: What does uncomment for commission mean?

Source/Experiments/SNOP/ECARun.m
```diff
@@ -0,0 +1,541 @@
...
+                [aSNOPModel startStandardRun:previousSR withVersion:previousSRVersion];
+                [[NSNotificationCenter defaultCenter] postNotificationName:ORECARunFinishedNotification object:self];
+                //Don't leak memory
+                [previousSR release];
+                [previousSRVersion release];
+            });
+        } else{
+            dispatch_sync(dispatch_get_main_queue(), ^{
+                [[NSNotificationCenter defaultCenter] postNotificationName:ORReleaseRunStateChangeWait object:self];
+                [aSNOPModel stopRun];
```
tlatorre-uchicago: In the normal case of a user hitting stop run, the run control will already be in the process of stopping a run and so this shouldn't be called. I guess there needs to be a distinction between the ECA thread normally exiting and when the run is stopped from the run control.

Source/Experiments/SNOP/ECARun.m
```diff
@@ -0,0 +1,541 @@
...
+        [tslope_delays_fine addObject:[NSNumber numberWithInt:0]];
+    }
+    
+    //EPED headers
+    [aSNOPModel updateEPEDStructWithCoarseDelay: 255 - floor(coarse_delay/10.0) fineDelay: fine_delay chargePulseAmp: 0x0	pedestalWidth: pedestal_width calType: 10 + (ECA_pattern+1)];
+    
+    //Inject pedestals
+    int eca_pattern_num_steps = getECAPatternNSteps(ECA_pattern);
+    NSMutableArray* pedestal_mask = getECAPattern(ECA_pattern);
+    if ([pedestal_mask count] == 0 || pedestal_mask == nil) {
```
tlatorre-uchicago: Should check nil first here. If it's not nil, even if the count is zero, it needs to be released.

Source/Experiments/SNOP/ECARun.m
```diff
@@ -0,0 +1,541 @@
...
+        NSLogColor([NSColor redColor], @"ECA: Error loading pedestal mask for pattern: %d \n", ECA_pattern);
+        return false;
```
tlatorre-uchicago: Need to release tslope_delays_coarse and tslope_delays_fine here.

Source/Experiments/SNOP/ECARun.m
```diff
@@ -0,0 +1,541 @@
...
+                dispatch_sync(dispatch_get_main_queue(), ^{
+                    [anMTCModel stopMTCPedestalsFixedRate];
+                });
+                
+                //Check if user has cancel ECAs and quit in that case
+                if([ECAThread isCancelled]) {
+                    NSLog(@"ECA cancelled by user... \n");
+                    //Don't leak memory
+                    [tslope_delays_coarse release];
+                    [tslope_delays_fine release];
```
tlatorre-uchicago: need to release pedestal mask here.

Source/Experiments/SNOP/ECARun.m
```diff
@@ -0,0 +1,541 @@
...
+                    return false;
+                }
+                
+            }
+            
+        } // end ECA steps
+    }
+    @catch(...){
+        //Don't leak memory
+        [pedestal_mask release];
```
tlatorre-uchicago: need to release tslope_delays_coarse and tslope_delays_fine here.

Source/Experiments/SNOP/ECARun.m
```diff
@@ -0,0 +1,541 @@
...
+        return false;
+    }
+    
+    [pedestal_mask release];
```
tlatorre-uchicago: and here.

Source/Experiments/SNOP/ECARun.m
```diff
@@ -0,0 +1,541 @@
...
+        //End of sub-run
+        dispatch_sync(dispatch_get_main_queue(), ^{
+            [anMTCModel stopMTCPedestalsFixedRate];
+        });
+        
+        //Check if user has cancel ECAs and quit in that case
+        if([ECAThread isCancelled]) {
+            NSLog(@"ECA cancelled by user... \n");
+            start_eca_run = FALSE;
+            return false;
```
tlatorre-uchicago: need to release pedestal_mask

Source/Experiments/SNOP/ECARun.m
```diff
@@ -0,0 +1,541 @@
...
+        }
+        
+    } // end ECA steps
+    
+    //OK run
```
tlatorre-uchicago: need to release pedestal_mask

Source/Experiments/SNOP/ECARun.m
tlatorre-uchicago: same comment as below. The nil check needs to be first, and if [pedestal_mask count] is 0 it still needs to be released.

Source/Experiments/SNOP/ECARun.m
```diff
@@ -0,0 +1,541 @@
...
+}
+
+- (BOOL) doPedestals
+{
+    
+    NSLog(@"Starting ECA Pedestal Run... \n");
+    
+    double coarse_delay = [anMTCModel dbFloatByIndex:8];
+    double fine_delay = [anMTCModel dbFloatByIndex:9];
+    double pedestal_width = [anMTCModel dbFloatByIndex:1];
```
tlatorre-uchicago: I don't think these are floats. Can you update the index to use the constants defined in ORMTC_Constants.h.

Source/Experiments/SNOP/ECARun.m
```diff
@@ -0,0 +1,541 @@
...
+}
+
+- (BOOL) doTSlopes
+{
+    
+    NSLog(@"Starting ECA TSlope Run... \n");
+    
+    double coarse_delay = [anMTCModel dbFloatByIndex:8];
+    double fine_delay = [anMTCModel dbFloatByIndex:9];
+    double pedestal_width = [anMTCModel dbFloatByIndex:1];
```
tlatorre-uchicago: same as above. I don't think these are floats, and you should use the defined indexes in ORMTC_Constants.h.

Source/Experiments/SNOP/ECARun.m
```diff
@@ -0,0 +1,541 @@
...
+                NSLog(@"************************* \n");
+                NSLog(@"ECA TSlope: STEP %d/%d - Point %d/%d \n",step+1,eca_pattern_num_steps, ipoint+1,tslope_nsteps);
+                NSLog(@"------------------------- \n");
+                NSLog(@" Coarse delay: %d\n",current_coarse_delay);
+                NSLog(@" Fine delay: %d\n",current_fine_delay);
+                NSLog(@"************************* \n");
+                
+                dispatch_sync(dispatch_get_main_queue(), ^{
+                    [anMTCModel setupGTCorseDelay:current_coarse_delay]; //UNCOMMENT FOR COMMISSION
+                    [anMTCModel setupGTFineDelay:current_fine_delay]; //UNCOMMENT FOR COMMISSION
```
tlatorre-uchicago: uncomment for commission?

Source/Experiments/SNOP/ECARun.m
tlatorre-uchicago: I just looked up the method and I see it just converts the value to a float so I guess this is OK, as long as they are later set as an unsigned short.

Source/Experiments/SNOP/ECARun.m
jcaravaca: These lines would make ORCA crash if running locally so I usually test the code with this lines commented. These notes are for me to remember to uncomment them before commissioning.

Source/Experiments/SNOP/ECARun.m
jcaravaca: If the model is not found the function will exit and still will be start_eca_run = FALSE, so the rest of the ECA functions won't run

Source/Experiments/SNOP/ECARun.m
jcaravaca: tslope_delays_coarse and tslope_delays_fine are set to autorelease and we are in an autorelease pool, so it should be automatically handled?

tlatorre-uchicago:
It looks like hvMasterTriggersOff is used to disable channel triggers. This will disable the channel triggers, but doesn't remember which crates had their triggers disabled. I think this needs to be implemented for a smooth transition back to a normal run.

The hvMasterTriggersOff will also not throw an exception if it fails to turn off triggers since it is done asynchronously. There is no easy way to do a blocking call to make sure that the channel triggers are successfully turned off in Orca right now. I could add such a method. Otherwise, the only option is the function:

```
- (void) loadHardwareWithSlotMask: (uint32_t) slotMask withCallback: (SEL) callback target: (id) self
```

which is a little complicated to use, especially if you are initializing all the crates.

jcaravaca:
Shouldn't we triggers enabled in all crates by default at the start of a run? At least that should happen for a physics run I assume.
Also I noticed that hvMasterTriggersOff doesn't return anything. It'd be convenient a method to know whether the triggers are actually off or not, not only for ECAs but in general.

Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
```diff
@@ -5268,6 +5274,10 @@ - (void) _hvXl3
 
 - (void) _setPedestalInParallelWorker
 {
+    
+    //Keep track of how many XL3s are changing ped mask
+    changingPedMask = TRUE;
```
tlatorre-uchicago: I think this whole function should be wrapped in a block like this:

```
@synchronized(self) {
    # function goes here
}
```

to make sure that two different threads don't overwrite the changingPedMask variable at the same time.

tlatorre-uchicago:
Well, crates will always be at low voltage some of the time, and we absolutely can't change that software lock by default anywhere because that is how the operator ensures that triggers are never enabled for a given crate.

I agree it would be nice to get a return value from that function. I will think about it and see how easy it is to implement.

I guess the best way to do this would be to not change the software lock because that's how the operator guarantees that triggers are off and it would be safest to never touch that. The alternative would be to:

1. save a copy of the triggers for every single FEC before starting the ECA run
2. disable the triggers (not via the software lock, but in the Fec32 model)
3. at the end of the ECA run reload the triggers to the FEC object
4. reload the hardware

That way the software triggers lock never needs to be changed from the ECA script and we don't have to worry about accidentally enabling triggers for a crate that the operator didn't intend to.

tlatorre-uchicago:
An easier but a bit hacky way of doing this would be:

- disable triggers for the crate but don't update the GUI
- call loadHardware at the end to reload the GUI settings

I actually like this method. We need a big warning though if the hardware load fails at the end of the ECA run so the operator knows that the GUI might be inconsistent with the HW state.

tlatorre-uchicago:
I have added two new functions to the XL3 model which should allow you to disable triggers during an ECA run and then reload them afterwards. The two functions are:

```
- (void) disableTriggers
- (void) loadTriggers
```

Both of them block and will throw an exception if there was an error when setting the channel triggers.

jcaravaca:
Great, thanks! I've added the disableTriggers method to the ECA algorithm. I assume that triggers are reloaded at the start of a run so I don't explicitly use the loadTriggers method when rolling over.

tlatorre-uchicago:
They are only reloaded on hard run starts, so I think they need to be reset after the ECA run finishes.

jcaravaca:
I call loadTriggers at the end of the ECAs and if it fails it will show a big warning and will not rollover to a new run. I think that makes sense because we don't want to start a physics run with half the triggers OFF.

Source/Experiments/SNOP/ECARun.m
tlatorre-uchicago: Still need to check nil first, i.e. if getECAPattern returns nil, the first check in the if statement will segfault. Is there a reason the ECA patterns can't just be defined as global variables and then you don't have to worry about checking for nil or releasing the memory?

Source/Experiments/SNOP/ECARun.m
tlatorre-uchicago: yes, you're right.

tlatorre-uchicago:
There is one last comment that needs addressing on checking for nil before accessing the pedestal_mask. After that is addressed, I will merge.

jcaravaca:
Last change in place, I'll test again at the teststand an let you know before merging.

jcaravaca:
Tested at the teststand 02/06

tlatorre-uchicago:
I just emailed Mark, and he says we are no longer required to support XCode 5.1! Woohoo!

Here are a couple of things that I just noticed when merging:

- [ ] There is a lot of whitespace on empty lines in this commit.
- [ ] The code needs to be updated to talk to the new MTC model which has just been merged.

Could you remove the whitespace? You can see it by running the following command:

```
$ git diff origin/master..HEAD
```

from your branch and the blank lines with whitespace should show up as red.

Eric is here on shift today, and so I can update the MTC model stuff while merging if you are OK with that. Let me know if that sounds good.

jcaravaca:
Removed trailing spaces.

Yeah maybe it's faster for you and Eric to do the merge. I'd like to have a look and test it at the teststand before definitely merging into master though, just in case.

marzece:
I made a branch that has the needed updates
https://github.com/marzece/orca/tree/eca 
Though there may be a few issues with units, I'm not sure. The coarse delay is in nano-seconds and the fine delay is now in pico seconds. Let me know if things seem good.

jcaravaca:
It looks good. I commented on a couple of things directly on his branch. I could merge @marzece branch, change the coarse and fine delays and push again to update the PR.

tlatorre-uchicago:
OK, that sounds good.

jcaravaca:
Merged with master. There are a couple of bugs related with the latest MTC PR that should be solved in PR #301. After merging that one we can test at LV.

tlatorre-uchicago:
merged.

Issue #246, opened 2017-01-17
Hardware wizard not properly setting channel triggers to off for open paddle cards


tannerbk:
If you try to turn on channel triggers for an entire crate using the hardware wizard, it should prevent you from turning them on for channels with a pulled resistor and channels with an open paddle card.

It seems to work for pulled resistors but not open paddle cards. I think @tlatorre-uchicago is already aware.

tlatorre-uchicago:
Fixed in #256.

PR #256, opened 2017-01-19
update HW Wizard to disable channel triggers if a channel is connected to an open HV relay

tlatorre-uchicago:
The HW wizard now checks the HV relays when deciding whether to enable or disable triggers for a particular channel.

With this change it should be possible to use the HW wizard to enable triggers for an entire crate and be sure that triggers are only enabled for channels actually at HV.


tlatorre-uchicago:
Tested on the teststand on Wednesday, January 18.

tlatorre-uchicago:
Fixes issue #246.

tlatorre-uchicago:
merged.

Issue #233, opened 2017-01-06
ORCA does not save itself automatically


fbdescamps:
We need it to save:
 * at the start of each run
 * every 5min

It was checked that ORCA does not currently save to file at the run start in this shift:
http://snopl.us/shift/view/b6886d7d0e9a4a92a4040242d0225698

fbdescamps:
@BenLand100 pointed out that this might be because of the fact that we use our own run control...

BenLand100:
That does look to be the case - see this line in the orca run control
https://github.com/snoplus/orca/blob/252a0bac2025b8d0bfd7cd3cc6fec6635a1181ae/Source/Objects/Control/RunControl/ORRunController.m#L666
which seems to be missing an analogous statement in our run control
https://github.com/snoplus/orca/blob/986b1630d03c4603be38c6714eb707903dab2210/Source/Experiments/SNOP/SNOPController.m#L411

fbdescamps:
@jcaravaca Any update on this?

fbdescamps:
@jcaravaca I think the latest here was that ORCA only saves to the current open file at run-start when the GUI actually changed. Has this been confirmed, and does ORCA indeed save at all types of run-starts when the GUI has changed? If so, please close this issue.

We will not be automatically saving every 5minutes or so due to the following arguments that were made against:
 * Experts sometimes use the unsaved state to be able to recover from changes that need to be reversed. This is a handy thing to keep for now.
 * If it done every run.. then you know the state. As in: ORCA crashes.. I wonder if that last change I  made was saved or not? That question is much easier if the save happens only at run boundaries.

tlatorre-uchicago:
Confirmed that ORCA does not save the .Orca file when you start a run from the SNO+ control panel.

fbdescamps:
Did you change something in the GUI before the run start?

tlatorre-uchicago:
Yes, we changed the PED -> PULGT button in the MTC GUI.

tlatorre-uchicago:
Fixed in #258.

PR #258, opened 2017-01-19
SNO+: update SNO+ run control to save document at run starts

tlatorre-uchicago:
This pull request updates the restart and resync buttons on the SNO+ run control to save the ORCA document before starting a new run.

Fixes issue #233.


tlatorre-uchicago:
Tested at the teststand on January 20, 2017.

tlatorre-uchicago:
merged.

PR #262, opened 2017-01-20
SNO+: Do not lock accept readback and rampdown for non-experts.

jcaravaca:
This required some changes in the HV lock logic.
It needs to be tested at the teststand and @BenLand100 approval.

### DIFF
--- a/Source/Experiments/SNOP/SNOPModel.m
+++ b/Source/Experiments/SNOP/SNOPModel.m
@@ -2010,6 +2010,16 @@ static NSComparisonResult compareXL3s(ORXL3Model *xl3_1, ORXL3Model *xl3_2, void
 
 }
 
+- (BOOL) isNotRunningOrInMaintenance
+{
+    
+    return (![gOrcaGlobals runInProgress] ||
+            (([gOrcaGlobals runType] & kMaintenanceRun) ||
+            ([gOrcaGlobals runType] & kDiagnosticRun)));
+    
+}
+
+
 @end
 
 @implementation SNOPDecoderForRHDR
--- a/Source/Objects/Custom Hardware/SNO+/XL3/XL3_LinkController.m
+++ b/Source/Objects/Custom Hardware/SNO+/XL3/XL3_LinkController.m
@@ -689,7 +689,8 @@ static NSDictionary* xl3Ops;
     }
     
     BOOL lockedOrNotRunningMaintenance = [gSecurity runInProgressButNotType:eMaintenanceRunType orIsLocked:ORXL3Lock];
-    BOOL notRunningOrInMaintenance = [aSNOPModel isNotRunningOrInMaintenance];
+    BOOL notRunningOrInMaintenance = true; //default if no snopmodel found
+    if (aSNOPModel) notRunningOrInMaintenance = [aSNOPModel isNotRunningOrInMaintenance];
     
     if ([hvPowerSupplyMatrix selectedColumn] == 0) { //A
         bool unlock = ![model hvANeedsUserIntervention] && [model hvEverUpdated] && [model hvSwitchEverUpdated] && ![model hvASwitch] && [model hvAFromDB] && !lockedOrNotRunningMaintenance;
--- a/Source/Objects/Custom Hardware/SNO+/XL3/XL3_LinkController.m
+++ b/Source/Objects/Custom Hardware/SNO+/XL3/XL3_LinkController.m
@@ -403,9 +403,10 @@ static NSDictionary* xl3Ops;
 #pragma mark •••Interface Management
 - (void) xl3LockChanged:(NSNotification*)aNotification
 {
+
     BOOL locked						= [gSecurity isLocked:ORXL3Lock];
     BOOL lockedOrNotRunningMaintenance = [gSecurity runInProgressButNotType:eMaintenanceRunType orIsLocked:ORXL3Lock];
-
+    
     //Basic
     [lockButton setState: locked];
     [basicReadButton setEnabled: !lockedOrNotRunningMaintenance];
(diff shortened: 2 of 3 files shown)


Source/Objects/Custom Hardware/SNO+/XL3/XL3_LinkController.m
```diff
@@ -707,24 +678,51 @@ - (void) monVltThresholdInInitChanged:(NSNotification*)aNote
...
-        [hvStepDownButton setEnabled:![model hvANeedsUserIntervention] && [model hvEverUpdated] && [model hvSwitchEverUpdated] && [model hvASwitch] && ![model hvARamping] && [model hvAFromDB]];
-        [hvRampUpButton setEnabled:![model hvANeedsUserIntervention] && [model hvEverUpdated] && [model hvSwitchEverUpdated] && [model hvASwitch] && ![model hvARamping] && [model hvAFromDB]];
-        [hvRampDownButton setEnabled:![model hvANeedsUserIntervention] && [model hvEverUpdated] && [model hvSwitchEverUpdated] && [model hvASwitch] && ![model hvARamping] && [model hvAFromDB]];
-        [hvStopRampButton setEnabled:![model hvANeedsUserIntervention] && [model hvEverUpdated] && [model hvSwitchEverUpdated] && [model hvASwitch] && [model hvARamping] && [model hvAFromDB]];
-        [hvAcceptReadbackButton setEnabled:[model hvANeedsUserIntervention] && [model hvAFromDB]];
+        bool unlock = ![model hvANeedsUserIntervention] && [model hvEverUpdated] && [model hvSwitchEverUpdated] && ![model hvASwitch] && [model hvAFromDB] && !lockedOrNotRunningMaintenance;
```
BenLand100: Take a closer look at the old code - it's not just the same logic duplicated each time though a few of them are the same. The dependence on [model hv*Ramping] was lost for ramp, step, and (not) stop  in the new version, and hvOn depends on ![model hv*Switch].

jcaravaca: Should be fixed now.

BenLand100: 👍

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -2042,6 +2042,26 @@ -(void) loadSettingsInHW
...
+    NSArray*  objs = [[(ORAppDelegate*)[NSApp delegate] document] collectObjectsOfClass:NSClassFromString(@"ORRunModel")];
+    ORRunModel* aRunControl;
+    if ([objs count]) {
+        aRunControl = [objs objectAtIndex:0];
+    } else {
+        NSLogColor([NSColor redColor], @"Couldn't find Run Control model. Please add it to the experiment and restart the run.\n");
+        return 0;
+    }
+
+    return (![aRunControl isRunning] ||
```
BenLand100: Happened to notice elsewhere some useful code for avoiding getting specific objects:
```
[gOrcaGlobals runInProgress] //true if in a run
[gOrcaGlobals runType] //returns the run type
```
`gOrcaGlobals` is defined in ORGlobal.h

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -2042,6 +2042,26 @@ -(void) loadSettingsInHW
...
+            (([aRunControl runType] & MAINTENANCE_RUN) ||
+            ([aRunControl runType] & DIAGNOSTIC_RUN)));
```
BenLand100: I was trying to write code similar to this for checking if we are in a physics run and realized (after an embarrassing amount of time) this needs to be updated to work with #235 (kMaintenanceRun, kDiagnosticRun)

jcaravaca:
Rebased with new master. This should be ready to go again.

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -2042,6 +2042,16 @@ -(void) loadSettingsInHW
 
 }
 
+- (BOOL) isNotRunningOrInMaintenance
+{
+    
+    return (![gOrcaGlobals runInProgress] ||
+            (([gOrcaGlobals runType] & kMaintenanceRun) ||
+            ([gOrcaGlobals runType] & kDiagnosticRun)));
```
tlatorre-uchicago: Can you confirm that this is the actual standard run and not what is selected in the SNO+ control panel?

tlatorre-uchicago:
Has this been tested on the teststand?

Source/Experiments/SNOP/SNOPModel.m
BenLand100: It's set when the ORRunStatusChangedNotification is fired ( https://github.com/snoplus/orca/blob/252a0bac2025b8d0bfd7cd3cc6fec6635a1181ae/Source/App%20Delegate/ORGlobal.m#L127 ) which happens in ORRunModel ( https://github.com/snoplus/orca/blob/252a0bac2025b8d0bfd7cd3cc6fec6635a1181ae/Source/Objects/Control/RunControl/ORRunModel.m#L860 ) latching the runType in that object. We set that runType in the loadStandardRun method ( https://github.com/snoplus/orca/blob/c7e261911b75c1343ab2028023105ab76d8bcab4/Source/Experiments/SNOP/SNOPModel.m#L1920 ) which is called prior to starting the run ( https://github.com/snoplus/orca/blob/c7e261911b75c1343ab2028023105ab76d8bcab4/Source/Experiments/SNOP/SNOPController.m#L455 ). It's pretty convoluted but tl;dr the result of that function is what's set by our run control and only updated when the run status changes.

tlatorre-uchicago: OK. Thanks!

jcaravaca:
Not yet. This requires HV so probably we want someone on site to be present while testing.

BenLand100:
I can test this and #264 tonight on the teststand.

BenLand100:
Tested on teststand 1/23 and confirmed this does fix #269

BenLand100:
Minor point of contention: these buttons are only not locked during maintenance runs and are still locked during other run types. This may be intentional to force operators to switch to a maintenance run before ramping down, but I was under the impression that we weren't going to lock them at all. If the latter is true this still needs modifications.

jcaravaca:
That was exactly the intention. I think we decided that'd be the strategy and in any case, we are most of the time running in Diagnostic and Maintenance currently. Thanks for testing this and confirming that it works fine.

BenLand100:
Ah just noticed some other things are no longer locked in this version that were previously (and probably should be). I needed to figure this out to test other stuff so see below:

The following are fixed by ORing in lockedOrNotRunningInMaininance here https://github.com/jcaravaca/orca/blob/fb6558786cb2156ce3e4f8b9a4da42965c97cd12/Source/Objects/Custom%20Hardware/SNO%2B/XL3/XL3_LinkController.m#L806 and changing this https://github.com/jcaravaca/orca/blob/fb6558786cb2156ce3e4f8b9a4da42965c97cd12/Source/Objects/Custom%20Hardware/SNO%2B/XL3/XL3_LinkController.m#L815 to a simple else.
* hvRelayMaskLowField
* hvRelayMaskHighField
* hvRelayMaskMatrix
* hvRelayOpenButton
* hvRelayCloseButton

And also hvTargetValueField needs to be locked after hvTargetValueStepper in updateHVButtons.

jcaravaca:
What about this? There were some conflicts with the OWL commit, so worth checking I didn't break any logic.

Source/Objects/Custom Hardware/SNO+/XL3/XL3_LinkController.m
```diff
@@ -707,24 +678,65 @@ - (void) monVltThresholdInInitChanged:(NSNotification*)aNote
...
 - (void) updateHVButtons
 {
+ 
+    //Get SNOP Model
+    NSArray*  objs = [[(ORAppDelegate*)[NSApp delegate] document] collectObjectsOfClass:NSClassFromString(@"SNOPModel")];
+    SNOPModel* aSNOPModel = nil;
+    if ([objs count]) {
+        aSNOPModel = [objs objectAtIndex:0];
+    } else {
+        NSLogColor([NSColor redColor], @"xl3LockChanged: Couldn't find SNOP model. Please add it to the experiment and restart the run.\n");
```
tlatorre-uchicago: Need to set some default value here and not access snopModel below. I'd say it's safest to assume it's not running or in maintenance mode.

BenLand100: Ugh yeah. Honestly it's probably better not to add a new method to the SNOPModel since everything is accessible through gOrcaGlobals and getting orca models is a pain. Let me see if I can push to Javi's branch.

BenLand100: On second thought the method may be useful elsewhere so fixed below.

marzece:
merged.

PR #263, opened 2017-01-21
SNO+: add ability to correct hv readback value with value from DB

BenLand100:
Correct readback by a multiplicative value readback_corr from the hvparams table on the ORCA side. 
I added this to both the detector and test databases (after confirming it did not break master) with the following query that allows for no more than a 10% correction defaulting to no correction:
```
alter table hvparams add column readback_corr float not null check (readback_corr >= 0.9 and readback_corr <= 1.1) default 1.0;
```

### DIFF
--- a/Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.h
+++ b/Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.h
@@ -122,6 +122,7 @@ enum {
     BOOL _isTriggerON;
     
     unsigned long _hvNominalVoltageA;
+    float _hvReadbackCorrA;
     float _hvramp_a_up;
     float _hvramp_a_down;
     float _vsetalarm_a_vtol;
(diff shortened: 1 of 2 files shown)


tlatorre-uchicago:
What do you think about putting the correction in the XL3 HV tab instead of printing it since it will always be the same value?

BenLand100:
I'd like to do GUI changes in a separate PR since there are multiple things I plan to change: some indication that relays for a crate are disabled due to 16B being on, showing nominals / other parameters at the bottom of the HV gui, a count of disabled relays, and things i'm forgetting off the top of my head.

tlatorre-uchicago:
Has this been tested?

BenLand100:
I only checked that it got the values from the DB not that the correction was applied correctly (though I would be surprised if it wasn't).

BenLand100:
Tested on teststand 1/23 - corrected readback was properly shown in logs, hv master, xl3 window, and programatically.

marzece:
merged.

PR #264, opened 2017-01-21
SNO+: alarm on setpoint changing during a physics run.

BenLand100:
The 80500 alarm block is now allocated for an edge alarm in the test and detector DBs for this case. It's a per-supply alarm like other HV alarms so we will have a record of which crate was changed. The alarm will fire whenever the runtype contains the physcs bit, but this is easy to change if we want other functionality. At the moment the what-to-do is "Do not do that. Start a new run instead."

Needs #262 to be fixed to unlock the HV controls to test on teststand.

(The removed if statement was redundant.)


BenLand100:
Tested on teststand 1/23 - worth noting that there is currently no way in the gui to change the setpoints during a physics run assuming all locks hold (I disabled them for testing), but better safe than sorry?

tlatorre-uchicago:
merged.

PR #271, opened 2017-01-24
SNO+: end editing of target value field prior to HV supply actions.

### LINKED ISSUE #226: If you hit ramp up while still editing the target field the ramp goes to the previous entered target
This might be intended behaviour but if you have the target set to (say) 2kV then change that to 200V then without hitting enter, you click ramp up, then the crate will ramp to 2kV. It might be better to have the ramp up button end the editing of the target field and use the value currently in there. Or just refuse to do any ramping if that field is still being edited.

BenLand100:
Ensures that a ramp will go to the value that appears to be typed into the text field (rounded to the nearest 12bit value) whether the user has remembered to press enter or not.

Coco idiosyncratic code stolen from https://stackoverflow.com/questions/2284149/forcing-a-cocoa-text-field-to-end-editing
Tested on teststand 1/23 (minor change). Fixes #226

### DIFF
--- a/Source/Objects/Custom Hardware/SNO+/XL3/XL3_LinkController.m
+++ b/Source/Objects/Custom Hardware/SNO+/XL3/XL3_LinkController.m
@@ -1402,11 +1402,13 @@ static NSDictionary* xl3Ops;
 
 - (IBAction)hvTurnOnAction:(id)sender
 {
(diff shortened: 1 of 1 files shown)


tlatorre-uchicago:
merged.

PR #273, opened 2017-01-24
HV GUI updates

BenLand100:
Displays database parameters at the bottom of the HV tab on the XL3 GUI, and adds a notification to OWL crates when the OWL supply (16B) is switched on. 

Here's a screenshot of the GUI so people can rate/hate:
hvguimods

This will need modification after #263 is merged (the FIXME)

### DIFF
--- a/Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
+++ b/Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
@@ -816,7 +816,7 @@ isLoaded = isLoaded;
 /**
  * Global variable to store the state of the OWL supply
  */
-BOOL owlSupplyState = true; //better safe than sorry
+BOOL owlSupplyState = false;
 
 /**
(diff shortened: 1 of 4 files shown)


Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
```diff
@@ -816,7 +816,7 @@ - (BOOL) hvEverUpdated
 /**
  * Global variable to store the state of the OWL supply
  */
-BOOL owlSupplyState = true; //better safe than sorry
+BOOL owlSupplyState = false;
```
BenLand100: I changed this to false because the default only matters in the case that the crate 16 XL3 is not connected and one would probably want relay control in that case / shouldn't assume the HV is on.

BenLand100:
@tlatorre-uchicago worth a quick look, but has been tested. If it looks good to you it's ready to merge.

tlatorre-uchicago:
merged.

PR #274, opened 2017-01-24
MTC development

marzece:
I'm submitting this PR a bit before things are totally good to go so that the review process can get started b/c I think things are somewhat close to their final state.

Here's a brief summary of all the changes here
- Removed a lot of the unused GUI elements
- Added DB communication for the MTCA thresholds and trigger scan
- Removed a lot of the register read/writes and replaces them with commands for the mtc-server
- General aesthetic and UI improvements 
- Removing the use of a "DB" in the mtc model
- Changing all the code in ELLIE and SNOPController that used the DB in the mtc model

Things that I'd still like to do before calling this ready to go. 
- [x] Make the threshold setting for when only a few trigger scans are available work better. Currently if any trigger scans are missing you can only set thresholds in DAC counts. I'd like to make it so you can set thresholds in NHIT on thresholds for which triggers scan are available while still locking the others out.
- [x] Get the "lock" picture to better positioned and tweak window sizes a bit more
- [x] I still need to figure out if/how this will work with the standard run saving and loading. I've largely not tested that. It may require a temporary "rosetta stone" version of orca that can load using the old keys and save using the new keys.
- [x] Remove use of cocoa bindings for a lot of the variables
- [x] Figure out how to coalesce notifications to avoid notification spam on start up.
- [x] Rebase this to master and make it mergable
- [x] Remove commit fc40a77 which I had to add to make it compile on my laptop

Also worth noting is there's a bug in the ORPQ result that's been fixed in the phil/detector_db branch ([link to commit)](https://github.com/snoplus/orca/commit/9987f04d1d47a6a2057cf063ef969d0b9a20989d) I haven't put it in this branch to avoid conflicts. But that fix will need to be in ORCA either here or from elsewhere before this branch will work.


marzece:
I think this is good to go. I have a version of this branch ([link](https://github.com/marzece/orca/tree/rosetta_whale)) that should help getting all the standard runs updated. I'll remove the pbx change once this has been reviewed.

jcaravaca:
@marzece the SR database can be pruned and the SR created from scratch if we not care about the current stored standard runs. That'd avoid any translation problem. In case we want to maintain those we can use your ORCA-rosseta.

Source/Experiments/SNOP/SNOPController.m
```diff
@@ -1872,117 +1859,33 @@ -(void) displayThresholdsFromDB
...
+            MTC_N100_HI_THRESHOLD_INDEX,
+            MTC_N100_MED_THRESHOLD_INDEX,
+            MTC_N100_LO_THRESHOLD_INDEX,
+            MTC_N20_THRESHOLD_INDEX,
+            MTC_N20LB_THRESHOLD_INDEX,
+            MTC_OWLN_THRESHOLD_INDEX,
+            MTC_ESUMH_THRESHOLD_INDEX,
+            MTC_ESUML_THRESHOLD_INDEX,
+            MTC_OWLEHI_THRESHOLD_INDEX,
+            MTC_OWLELO_THRESHOLD_INDEX};
```
tlatorre-uchicago: This is defined twice. Maybe make it a global variable?

Source/Objects/Custom Hardware/SNO+/MTC/ORMTCModel.m
```diff
@@ -855,48 +698,238 @@ - (void)encodeWithCoder:(NSCoder*)encoder
...
+    if(in_units == MTC_RAW_UNITS) {
+        float value_in_mv = ((aThreshold - [self getBaselineOfType:type])/DAC_per_mv);
+        
+        if(out_units == MTC_mV_UNITS)
+        {
+            return value_in_mv;
+        }
+        else if (out_units == MTC_NHIT_UNITS)
+        {
+            return [self convertThreshold:value_in_mv OfType:type fromUnits:MTC_mV_UNITS toUnits:out_units];
```
tlatorre-uchicago: Why does this need to recursively call itself? Shouldn't DAC -> NHIT just use the trigger scan directly?

This function seems a little complicated. I wonder if it would be easier to understand if there were just a few functions like:

```
- (float) convertADCToNhit: (uint16_t) adcValue forTriggerType: (int) type;
- (uint16_t) convertNhitToADC: (float) nhit forTriggerType: (int) type;
- (float) convertADCToMv: (uint16_t) adcValue forTriggerType: (int) type;
```
etc.

Source/Objects/Custom Hardware/SNO+/MTC/ORMTCModel.m
```diff
@@ -855,48 +698,238 @@ - (void)encodeWithCoder:(NSCoder*)encoder
...
-		NSDecimalNumber* aValue = [NSDecimalNumber decimalNumberWithString:anObject];
-		[mtcDataBase setObject:aValue forNestedKey:[self getDBKeyByIndex:anIndex]];
-	}
-	else {
-		[mtcDataBase setObject:anObject forNestedKey:[self getDBKeyByIndex:anIndex]];
-	}
+- (uint16_t) getBaselineOfType:(int) type {
+    return mtca_baselines[type];
+}
+- (void) setBaselineOfType:(int) type toValue:(uint16_t) _val {
```
tlatorre-uchicago: Maybe change Type -> Trigger?

Source/Objects/Custom Hardware/SNO+/MTC/ORMTCModel.m
```diff
@@ -855,48 +698,238 @@ - (void)encodeWithCoder:(NSCoder*)encoder
...
+        NSNotification* note = [NSNotification notificationWithName:ORMTCABaselineChanged object:self];
+        [[NSNotificationQueue defaultQueue] enqueueNotification:note postingStyle:NSPostASAP coalesceMask:NSNotificationCoalescingOnName forModes:nil];
+    }
+}
 
-    [[NSNotificationCenter defaultCenter] postNotificationName:ORMTCModelMtcDataBaseChanged object:self];
+- (float) DAC_per_NHIT_ofType:(int) type {
+    return mtca_dac_per_nhit[type];
+}
+- (void) setDAC_per_NHIT_OfType:(int) type toValue:(float) _val {
```
tlatorre-uchicago: Since all of the other functions in the model are camel case, you should make these be camel case too.

Source/Objects/Custom Hardware/SNO+/MTC/ORMTCModel.m
```diff
@@ -855,48 +698,238 @@ - (void)encodeWithCoder:(NSCoder*)encoder
...
+}
+- (void) setConversionIsValidForThreshold:(int) type isValid:(BOOL) _val {
+    if(mtca_conversion_is_valid[type] != _val) {
+        mtca_conversion_is_valid[type] = _val;
+        NSNotification* note = [NSNotification notificationWithName:ORMTCAConversionChanged object:self];
+        [[NSNotificationQueue defaultQueue] enqueueNotification:note postingStyle:NSPostASAP coalesceMask:NSNotificationCoalescingOnName forModes:nil];
+    }
+}
+
+- (NSString*) StringForThreshold:(int) threshold_index {
```
tlatorre-uchicago: most functions in the MTC model start with a lower case letter.

Source/Objects/Custom Hardware/SNO+/MTC/ORMTCModel.m
```diff
@@ -1114,7 +1077,7 @@ - (void) setSingleGTWordMask:(unsigned long) gtWordMask
 {	
 	@try {
 		[self setBits:kMtcMaskReg mask:gtWordMask];
-		NSLog(@"Set GT Mask: 0x%08x\n",uLongDBValue(kGtMask));
+        //NSLog(@"Set GT Mask: 0x%08x\n",uLongDBValue(kGtMask)); This isn't right fixit
```
tlatorre-uchicago: ?

Source/Objects/Custom Hardware/SNO+/MTC/ORMTCModel.m
```diff
@@ -1128,7 +1091,7 @@ - (void) clearSingleGTWordMask:(unsigned long) gtWordMask
 	@try {
 		//[self clrBits:kMtcGmskReg mask:gtWordMask];
 		[self clrBits:kMtcMaskReg mask:gtWordMask];
-		NSLog(@"Set GT Mask: 0x%08x\n",uLongDBValue(kGtMask));
+        //NSLog(@"Set GT Mask: 0x%08x\n",uLongDBValue(kGtMask)); This isn't right fixit
```
tlatorre-uchicago: ?

Source/Objects/Custom Hardware/SNO+/MTC/ORMTCModel.m
```diff
@@ -1788,26 +1495,33 @@ - (void) basicMTCReset
...
-
+    int server_index;
+    for(i=FIRST_MTC_THRESHOLD_INDEX;i<=LAST_MTC_THRESHOLD_INDEX;i++)
+    {
+        @try {
+            server_index = [self model_index_to_server_index:i];
+            dacs[server_index] = [self getThresholdOfType:i inUnits:MTC_RAW_UNITS];
+            [self validateMTCADAC:dacs[server_index]];
+        } @catch (NSException* excep) {
+            @throw; //Let it bubble up
```
tlatorre-uchicago: Is there an advantage to catching and then throwing the exception here?

tlatorre-uchicago:
It looks like these changes removed the ability to set individually the coarse and fine delays. I think this is something that ECA might need, but you should check with @jcaravaca .

Source/Objects/Custom Hardware/SNO+/MTC/ORMTCModel.m
```diff
@@ -1462,152 +1244,73 @@ - (void) setTheLockoutWidth:(unsigned short) theLockoutWidthValue
...
 	@try {
-		[self setBits:kMtcControlReg mask:MTC_CSR_PULSE_EN];
+        [mtc okCommand:"enable_pulser"];
 		NSLog(@"Enabled Pulser.\n");		
 	}
 	@catch(NSException* localException) {
 		NSLog(@"Unable to enable the pulser!\n");		
-		[localException raise];	
+		[localException raise];
+        return;
```
tlatorre-uchicago: don't need a return in void function.

Source/Objects/Custom Hardware/SNO+/MTC/ORMTCModel.m
```diff
@@ -1616,12 +1319,13 @@ - (void) enablePulser
...
-		[self clrBits:kMtcControlReg mask:MTC_CSR_PULSE_EN];
-		NSLog(@"Disabled Pulser.\n");		
+        [mtc okCommand:"disable_pulser"];
+        NSLog(@"Disabled Pulser.\n");
 	}
 	@catch(NSException* localException) {
 		NSLog(@"Unable to disable the pulser!\n");		
-		[localException raise];	
+		[localException raise];
+        return;
```
tlatorre-uchicago: same here.

Source/Objects/Custom Hardware/SNO+/MTC/ORMTCModel.m
```diff
@@ -1630,24 +1334,26 @@ - (void) disablePulser
...
 	@try {
-		[self setBits:kMtcControlReg mask:MTC_CSR_PED_EN];
+        [mtc okCommand:"enable_pedestals"];
 		NSLog(@"Enabled Pedestals.\n");		
 	}
 	@catch(NSException* localException) {
 		NSLog(@"Unable to enable the Pedestals!\n");		
-		[localException raise];	
+		[localException raise];
+        return;
```
tlatorre-uchicago: and here.

Source/Objects/Custom Hardware/SNO+/MTC/ORMTCModel.m
```diff
@@ -1630,24 +1334,26 @@ - (void) disablePulser
...
-		[self clrBits:kMtcControlReg mask:MTC_CSR_PED_EN];
-		NSLog(@"Disabled Pedestals.\n");		
+        [mtc okCommand:"disable_pedestals"];
+		NSLog(@"Disabled Pedestals.\n");
 	}
 	@catch(NSException* localException) {
 		NSLog(@"Unable to disable the Pedestals!\n");		
-		[localException raise];	
+		[localException raise];
+        return;
```
tlatorre-uchicago: and here.

Source/Objects/Custom Hardware/SNO+/MTC/ORMTCModel.m
```diff
@@ -1788,26 +1495,33 @@ - (void) basicMTCReset
 	}
 }
 
+- (void) validateMTCADAC:(uint16_t) dac_value {
+    if( dac_value < 0 ) {
```
tlatorre-uchicago: don't need this check since it's a uint16_t.

Source/Experiments/SNOP/SNOPController.m
```diff
@@ -1476,95 +1495,119 @@ - (IBAction)standardRunNewValueAction:(id)sender
...
+            [mtcModel setPgt_rate:raw];
+            return;
+            break;
     }
-    //Prescale
-    if(activeCell == 10) {
-        raw = [[sender cellAtRow:10 column:0] floatValue];
-        [mtcModel setDbFloat: raw forIndex:kNhit100LoPrescale];
+    @try{
+        if(![mtcModel ConversionIsValidForThreshold:threshold_index])
```
tlatorre-uchicago: This is a bit weird. So, does this assume that if a model conversion isn't valid that the value in the GUI is raw instead of NHIT? It looks like this is also being called for prescale and the pulser. Is that right?

Source/Experiments/SNOP/SNOPController.m
```diff
@@ -1476,95 +1495,119 @@ - (IBAction)standardRunNewValueAction:(id)sender
...
-        raw = [[sender cellAtRow:11 column:0] floatValue];
-        [mtcModel setDbFloat: raw forIndex:kPulserPeriod];
+    @catch(NSException *excep) {
+        NSLogColor([NSColor redColor], @"Error while trying to set the MTC threshold, reason: %@\n",[excep reason]);
     }
 }
+-(void) updateThresholdDisplayAt:(int) row forThreshold:(int) type isNHit:(BOOL) isNhit isInMask:(BOOL) inMask usingModel:(id) mtcModel andFormatter:(NSFormatter*) formatter
+{
+    float value;
+    [formatter retain]; // Not sure if necessary, but shouldn't hurt.
```
tlatorre-uchicago: This is not necessary since the GC does not run in the middle of functions.

Source/Experiments/SNOP/SNOPController.m
tlatorre-uchicago: It looks like below you set the label depending on the units. It's probably better to check that or to maintain that state somewhere else.

jcaravaca:
We definitely need to set up the coarse and fine delays for ECAs

Source/Objects/Custom Hardware/SNO+/MTC/ORMTCController.m
```diff
@@ -839,313 +709,540 @@ - (IBAction) standardSetPedestalsDelay:(id) sender
...
-- (IBAction) standardStopFindTriggerZeroes:(id) sender 
-{
-	[self buttonPushed:sender];
-}
 
 - (IBAction) standardPulserFeeds:(id)sender
 {
-    [model setIsPedestalEnabledInCSR:[[sender selectedCell] tag]];
+    [model setIsPedestalEnabledInCSR:[includePedestalsCheckBox state]];
+    [self endEditing];
```
marzece: If you're editing a field (say the pulser rate) when the pulser feed button is clicked it will end the editing in that field the same way as if you'd hammered "enter" after typing that value.

tlatorre-uchicago: Ah OK. And it does that for all of the fields?

Source/Objects/Custom Hardware/SNO+/MTC/ORMTCController.m
```diff
@@ -839,313 +709,540 @@ - (IBAction) standardSetPedestalsDelay:(id) sender
...
-            [model setDefaultFile:[[openPanel URL]path]];
-            NSLog(@"MTC DataBase default file set to: %@\n",[[[[[openPanel URLs] objectAtIndex:0]path]stringByAbbreviatingWithTildeInPath] stringByDeletingPathExtension]);
+    int numRows, numCols;
+    int threshold_index;
+    NSString* name =nil;
+    NSString* baseline=nil;
+    NSString* dac_per_nhit=nil;
+    @try {
+        if (!result) {
+            [NSException raise:@"MTCControllerError" format:@"NULL result returned from database"];
```
marzece: Since this function is called when the MTC gui is "woken up" that'd mean if you'd get an error pop up when you start up orca. I think that'd be annoying, maybe not prohibitively so though. Also there kind of is a GUI affect if the DB isn't available. The MTC threshold dialog won't let you edit in non-raw units if the trigger scan isn't retrieved, and the SNOPController GUI will display all thresholds in raw units.

But if people think getting an annoying warning when you start up orca with the MTC and the DB isn't available is tolerable I can try and add it.

Source/Objects/Custom Hardware/SNO+/MTC/ORMTCController.m
```diff
@@ -839,313 +709,540 @@ - (IBAction) standardSetPedestalsDelay:(id) sender
...
+        if (numRows <= 0) {
+            [NSException raise:@"MTCControllerError" format:@"No rows returned from database"];
+        }
+        
+        if (numCols != 3) {
+            [NSException raise:@"MTCControllerError" format:@"Unexpected number of columns returned from database. Expected 3 got %i", numCols];
         }
-    }];
+    }
+    @catch (NSException *exception) {
```
marzece: Think of it as a `goto err;` type statement. 

It seems that using exception handling as a flow control type things is a bit of a hotly debated thing. It seems that the strongest argument against it is that it incurs more system overhead than normal validation. Which I don't think is too concerning for this part of code since it will be called ~once a day.

Anyways, I'll change it to just use return statements to bomb out early. But, I don't think there's anything morally wrong with the code as it is.

Source/Objects/Custom Hardware/SNO+/MTC/ORMTCModel.m
marzece: The recursive call is needed b/c I only do the baseline subtraction when going between millivolts and raw units. I remember thinking it through why that had to be the case, and I think it was basically b/c the 'baseline' in nhit is defined to be zero. So you can only really do the subtraction by going through millivolts first.

And I think that's why I determined it was worth doing it all in one (recursive) function rather than in separate functions. That way I only have to define the baseline subtraction in exactly one place.

I'm sure all that could be done with separate functions as well. I'll think about doing it...but I kinda don't really want to.

Source/Objects/Custom Hardware/SNO+/MTC/ORMTCModel.m
marzece: Readability, to make it clear I was explicitly not catching exceptions. I could remove the try catch and just add a comment though

Source/Objects/Custom Hardware/SNO+/MTC/ORMTCController.m
```diff
@@ -236,15 +248,19 @@ - (void) updateWindow
...
+    [self mtcPrescaleValueChanged:nil];
 	[self isPulserFixedRateChanged:nil];
 	[self fixedPulserRateCountChanged:nil];
 	[self fixedPulserRateDelayChanged:nil];
     [self triggerMTCAMaskChanged:nil];
     [self isPedestalEnabledInCSRChanged:nil];
+    [self displayMasks];
+    [lockOutWidthField setIntValue:[model lockoutWidth]];
+    [pedestalWidthField setIntValue:[model pedestalWidth]];
+    [pedDelayField setIntValue:[model pedestalDelay]];
```
marzece: Those don't have any notification associated with them. The other fields are all updated by calls to the functions that handle their associated notifications.

Source/Objects/Custom Hardware/SNO+/MTC/ORMTCController.m
marzece: I believe so yes. All fields in the MTC Controller that is

Source/Objects/Custom Hardware/SNO+/MTC/ORMTCModel.m
```diff
@@ -1708,7 +1750,7 @@ - (void) setupDefaults
     [self setLockoutWidth:420];
     [self setPedestalWidth:52];
     [self setPrescaleValue:1];
-    [self setPgt_rate:10];
+    [self setPgtRate:10];
     [self setPedestalDelay:60];
```
jcaravaca: I'm confused about the pedestal delay. Is this the old 'internal delay'? If not, how is the latter handled? That one used to be set to 18ns and I think it shouldn't be touched, otherwise it messes up with the time calibration.

Source/Objects/Custom Hardware/SNO+/MTC/ORMTCModel.m
```diff
@@ -1287,9 +1302,43 @@ - (void) loadPedestalDelayToHardware
...
+    // Update the model variables apropriately
+    float current_delay = [self pedestalDelay];
+    uint16_t current_coarse_delay = (int)current_delay % 10;
+    [self setPedestalDelay:current_delay-current_coarse_delay+coarse_delay];
+}
+
+- (void) loadFineDelayToHardware: (float) fine_delay {
+    @try {
+        [mtc okCommand:"set_fine_delay %f", fine_delay];
+        NSLog(@"mtc: coarse delay %.2fns\n", fine_delay);
```
jcaravaca: should be 'fine delay'

Source/Objects/Custom Hardware/SNO+/MTC/ORMTCModel.m
```diff
@@ -1287,9 +1302,43 @@ - (void) loadPedestalDelayToHardware
...
+
+    }
+    @catch(NSException* localException) {
+        NSLog(@"Could not set coarse delay\n");
```
jcaravaca: same

Source/Objects/Custom Hardware/SNO+/MTC/ORMTCController.m
```diff
@@ -749,48 +644,32 @@ - (IBAction) basicStopAction:(id) sender
...
+        NSLogColor([NSColor redColor], @"Error loading the MTCA DACs. Reason: %@\n.",[excep reason]);
+    }
 }
 
-- (IBAction) standardSetCoarseDelay:(id) sender 
+- (IBAction) setAdvancedOptions:(id)sender
 {
-	[model setupGTCorseDelay];
+    @try{
+        [model loadPedestalDelayToHardware];
```
jcaravaca: shouldn't you call loadCoarseDelayToHardware and loadFineDelayToHardware here as well?

marzece:
@jcaravaca I added two functions loadCoarseDelayToHardware and loadFineDelayToHardware in 049cc32. Let me know if those should suffice.

Source/Objects/Custom Hardware/SNO+/MTC/ORMTCController.m
marzece: No, the MTCModel largely acts as a pass through to the mtc-server. So for normal setting of the ped delay it just sends a similar request to the server (`set_gt_delay delay`). The server then handles splitting the delay into its coarse and fine components.  For the new functions that specifically set the coarse/fine delay they call a similarly named server command (`set_coarse_delay delay` or `set_fine_delay delay`).

jcaravaca: (same comment as above, probably missed with the changes...) What about the old 'internal delay'? Is it always zero? That one used to be set to 18ns and I think it shouldn't be touched, otherwise it messes up with the time calibration.

tlatorre-uchicago: The internal delay cannot be changed. It is in the hardware and I think it is 18.35 ns.

jcaravaca: Then I'm confused by the internal delay field 'MinDelayOffset' of the old MTC model. Was that useless then?

tlatorre-uchicago: Yeah, Phil and I just looked at that yesterday and it is used nowhere in the MTC model.

tlatorre-uchicago:
I've merged phil's detector db branch. Would you mind merging and making sure the detector state changes are set correctly in the MTC model?

marzece:
I updated for Phil's detector db changes. However I couldn't figure out how to convert between values for the PED delay from the DB and how I use them. I'll have to consult with you (Tony) again to get it right. Other than that everything is tested and certainly seems to work (with the exception of the MTCA relays which I can't test on the teststand.)

tlatorre-uchicago:
merged.

PR #276, opened 2017-01-25
Handle case when database can't be connected to 

EdLeming:
Avoids nasty seg faults and attempts to tell the user what's happening.


tlatorre-uchicago:
merged.

PR #286, opened 2017-01-31
SNO+: update readHVStatus to prompt user to disable triggers if a crate is at low voltage

tlatorre-uchicago:
This commit adds a safety check which prompts the user to disable triggers if
the first HV readback from the XL3 is not at nominal voltage and the triggers
are enabled. This will hopefully prevent enabling triggers at low voltage if an
operator accidentally opens an old Orca file.


tlatorre-uchicago:
Tested on the teststand on January 31, 2017.

marzece:
Perhaps this should not be only for the first read. Last night we had two power supply trips and we had to remember to turn the triggers off both times since the crate was no longer at HV.

tlatorre-uchicago:
@marzece I opened a new pull request #293.

tlatorre-uchicago:
merged.

PR #293, opened 2017-02-01
SNO+: prompt the operator to disable triggers after a HV setpoint discrepancy

tlatorre-uchicago:
A set point discrepancy most likely means the supply tripped which means we should prompt the user to turn triggers off for that crate as soon as possible.

@BenLand100 Can you have a look at this?

It's getting a bit crowded in this function but I couldn't find a better place to put it.


BenLand100:
It's worth noting this only applies to supply A trips, but otherwise looks fine. Would we want something similar for the owls or do we not worry about those as much since there's only one slot per crate with OWLS?

tlatorre-uchicago:
That's right. I believe Josh said an entire crate with triggers going could generate 2W of power. So, for a single slot that should be 125 mW. I'm not 100% sure at what level the transistor gets burned, but I suspect it's much less likely with one card, and given the complexity of dealing with that I didn't want to implement it.

tlatorre-uchicago:
Tested on the teststand on February 1, 2017.

tlatorre-uchicago:
merged.

PR #295, opened 2017-02-01
XL3 multi set triggers and sequencers

tlatorre-uchicago:
This pull request adds several functions to the XL3 model which allow one to set just the channel triggers and sequencers independently of loading the rest of the crate settings. It also adds a new function `disableTriggers` to the XL3 model which has the following properties:

- doesn't update the GUI
- raises an exception if there was an error
- blocks until the settings have been loaded

This function will be useful for the ECA run since it will ideally want to disable channel triggers, run the ECA run and then reload all the channel trigger settings afterwards.


tlatorre-uchicago:
Tested on the teststand on February 1, 2017.

tlatorre-uchicago:
merged.

PR #296, opened 2017-02-02
SNO+: decouple HV target value from ramping logic.

### LINKED ISSUE #285: HV target field value should be updated to the setpoint on start
This is true for the backend code in the model but is not propagated to the GUI. Perhaps makes the most sense to only ensure the target >= setpoint and target <= nominal for the GUI.

BenLand100:
If the readback on start is higher than the saved target value also
update the target value (and the GUI as a consequence) to the
readback to ensure that the target is always >= the setpoint. This
fixes a popup message informing the user that the target must be
greater than the setpoint upon first switching to the XL3 window
after opening a .orca file with a target saved below the readback.

Removed undo manager calls since methods will be called from HV thread and no need for this to be undoable. Fixes #285.


tlatorre-uchicago:
Does the target variable not also get set by the user when they change the field in which case we should preserve the ability to undo it?

BenLand100:
It does get set when the target value field is changed from the GUI, but I don't really see the advantage of having this operation be undo able in the first place. It's not as if the value previously in the target field was in any way special and in need of recovery if accidentally changed, one would just set their desired target and ramp up. It would be easy enough to async dispatch the undo to the main thread if you think it's a desirable feature though - let me know.

tlatorre-uchicago:
I agree it's unlikely to be useful, but I think it would be good to keep it that way for two reasons:

1. It maintains consistency across the GUI

2. If a operator for whatever reason decides they want to undo the HV target field then when they press Command-Z it will undo a previous action potentially hidden in another window which would be very confusing.

Is it possible to only display this warning when the user hits "Ramp UP"?

I think there are a couple of situations which might lead to potential confusion: if an XL3 disconnects/reconnects and reports some funny value this will lead to the target value potentially changing. During a handover shift this might cause the target point to increase due to the readback not being perfect.

Maybe a solution is to have the detector state readback set the target value. Is the target value stored in dac counts?

BenLand100:
@tlatorre-uchicago I added back the undo manager lines for the target value. After thinking about this a bit I decided the best solution was to separate the target value from the ramping so that there was no constraint that the target must be greater than the setpoint. This allows us to 1) remove the warning that caused #285 and 2) "Ramp to Target" instead of "Ramp Up" since there's no longer an issue with targets below the setpoint.

Haven't tested this on teststand yet.

BenLand100:
@tlatorre-uchicago tested on the teststand just now - there were some issues to resolve but this is now good to go for the next release.

tlatorre-uchicago:
merged.

PR #299, opened 2017-02-06
Remove uncatchable exception in selector sequence

marzece:
I asked Mark Howe about this one. He said it should be okay


tlatorre-uchicago:
merged.

Issue #300, opened 2017-02-07
caen initBoard function doesn't write acquisition mode


tlatorre-uchicago:
In initBoard, the CAEN model writes the acquisition mode before doing a software reset and so the first two bytes and the last bit of the acquisition control register are cleared and never reset.

tlatorre-uchicago:
Fixed in #306.

PR #306, opened 2017-02-08
SNO+: fix a bug in CAEN initialization

tlatorre-uchicago:
This commit fixes a bug in the initBoard function of the CAEN model.
Previously, the acquisition control register was written before performing a
software reset, which causes all of the registers to be set to their default
values, thus overwriting the settings in the acquisition control register. Now,
the software reset is performed first and the acquisition control register
written afterwards.

This fixes issue #300.


tlatorre-uchicago:
Tested at the teststand on February 14, 2017.

tlatorre-uchicago:
merged.

PR #308, opened 2017-02-08
SNO+: Check if XL3 is connected when dis/enabling triggers

jcaravaca:


tlatorre-uchicago:
merged.

PR #309, opened 2017-02-09
	SNO+: Don't disable pedestal when stopping pulser

boardhead:
The pedestal state in the detector db will now be valid.
Requires testing on hardware.
Also removed other TO_DO comments which are now done.


tlatorre-uchicago:
merged.

PR #311, opened 2017-02-09
	SNO+: Read MB and DC ID's from detector database

boardhead:
Sets ORCA FEC32 and DaughterCard ID's from detector database when loaded.
Tested without hardware and it seems to work.


tlatorre-uchicago:
merged.

PR #314, opened 2017-02-13
SNO+: update clearGlobalTriggerWordMask in MTC model

tlatorre-uchicago:
The clearGlobalTriggerWordMask in the MTC model will now clear the global
trigger mask without updating the GUI.

This fixes a bug in the recent MTC updates which caused a new run to clear the GT mask on every run start.


tlatorre-uchicago:
Tested on the teststand on February 14, 2017.

tlatorre-uchicago:
merged.

Issue #317, opened 2017-02-17
tubii caen channel select is wrong


tlatorre-uchicago:
On the Analog tab of the TUBii window, the CAEN channel select is backwards.

tlatorre-uchicago:
Fixed in #319.

PR #319, opened 2017-02-17
Short fix to the CAEN channel select buttons in the TUBii window.

icoulter:
Should fix issue #317 
@tlatorre-uchicago


tlatorre-uchicago:
merged.

PR #320, opened 2017-02-17
SNO+: Force load MTC settings in HW for ECA runs

jcaravaca:
Fixed some issues found on the fly during UG testing. Mainly related with the new MTC model.


jcaravaca:
@tlatorre-uchicago it'd be nice if we could merge this PR. This was tested at the detector so it should be good to go. In this way we can start taking ECAs asap.

tlatorre-uchicago:
Sure, we'll have to wait for the new release too though. There are a few other changes that need to make it in, but I'll try to get that done by the end of the week.

tlatorre-uchicago:
merged.

PR #323, opened 2017-02-20
Remove window duplication for SNOP dialogs

### LINKED ISSUE #291: CTRL-C on any window makes a copy of it
Probably related with #253
image

marzece:
Fixes #291, which was already closed. But I think we can agree it's best not to have two dialogs when there really shouldn't ever be for most SNO+ things. 

We'll probably have to get Mark Howe approval for this one, but it's probably worth having others here look at it first.

### DIFF
--- a/Source/Experiments/SNOP/SNOPController.h
+++ b/Source/Experiments/SNOP/SNOPController.h
@@ -178,6 +178,7 @@
 
 - (void) updateSettings: (NSNotification *) aNote;
 - (void) initializeUnits;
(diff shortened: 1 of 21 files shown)


tlatorre-uchicago:
I think there are some use cases for having two windows open. In the case of the SNO+ control panel it is useful to be able to see both the HV master tab and the standard run settings. It does mean we have to be extra careful to make sure that all of the GUI state is stored in the model and that the controller only syncs the model state to the GUI to make sure that they are never inconsistent.

PR #325, opened 2017-02-21
SMELLIE overhead reduction

EdLeming:
During testing last week it was noted that the overheads between smellie pulse sequences were larger than expected. The problem has since been identified and a solution has been implemented in the smellie server software. This PR adds minor changes to  the ellie model to facilitate those changes.


tlatorre-uchicago:
merged.

PR #328, opened 2017-02-23
Change database read logic to use ORCouchDB class

EdLeming:
This PR is addresses issue #327. Crashes were happening from inside blocks being placed into global queues e.g., via:

dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^ {
    // function call
});

I was using two instances of this syntax to load both the TELLIE and SMELLIE defaults from couch. If you changed the queue priorities, whichever one was on the lower priority would cause a crash. I'm not sure exactly why this is, but it seems running synchronous NSURLRequests on these dispatch queues spawns some extra threads, stressing the GCD queuing. I guess some other request has been pulled in since our testing which is pushing these queues a little harder, hence why we didn't observe during testing.

The solution is to use the ORCouchDB class written by M. Howe. This formats asynchronous requests into its own queue. I've had to change some of the logic flow to handle this, but it's essentially tidied up the code quite a lot. In going through this way I've also found a lot of old functions from Chris which are now defunct. I've removed these where ever I've found them.


tlatorre-uchicago:
Have you gotten a chance to test this?

tlatorre-uchicago:
I'm getting the following warnings when compiling this:

- Method definition for 'fetchSmellieConfigurationInformation' not found
- Method definition for 'smellieDBpush:' not found
- Method definition for 'smellieConfigurationDBpush:' not found

EdLeming:
I haven't tested UG. I'll try to get some time with the Friday morning shift.

tlatorre-uchicago:
merged.

PR #329, opened 2017-02-25
SNO+: Fix supply B curent near zero bug.

BenLand100:
Was using supply A's minimum value for the current limit, hence the spurious alarms.


tlatorre-uchicago:
merged.

PR #331, opened 2017-02-25
SNO+: New features for standard runs

jcaravaca:
- Store SR settings in model so that we only need one query
- Hardcode DIAGNOSTIC run to be available with no DB access
- Save current run type and version in .Orca file
- Move most of the SR logic from the controller to the model
- Display current run type and version only if run actually started
- Never lock SR tab
- Load SR at RESYNC

### DIFF
--- a/Source/Experiments/SNOP/SNOPModel.m
+++ b/Source/Experiments/SNOP/SNOPModel.m
@@ -64,6 +64,7 @@ NSString* ORSNOPModelDebugDBIPAddressChanged = @"ORSNOPModelDebugDBIPAddressChan
 NSString* ORSNOPRunTypeWordChangedNotification = @"ORSNOPRunTypeWordChangedNotification";
 NSString* ORSNOPRunTypeChangedNotification = @"ORSNOPRunTypeChangedNotification";
 NSString* ORSNOPRunsLockNotification = @"ORSNOPRunsLockNotification";
+NSString* ORSNOPModelSRCollectionChangedNotification = @"ORSNOPModelSRCollectionChangedNotification";
 NSString* ORSNOPModelSRChangedNotification = @"ORSNOPModelSRChangedNotification";
 NSString* ORSNOPModelSRVersionChangedNotification = @"ORSNOPModelSRVersionChangedNotification";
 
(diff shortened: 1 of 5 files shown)


jcaravaca:
Tested at the teststand 2/24

tlatorre-uchicago:
Just tested this on the teststand, and the DIAGNOSTIC run does not seem to come up without DB access.

jcaravaca:
One of the notifications was missing. It was only working for the cases in which the settings where not well introduced but the server was running. It should be good now.

tlatorre-uchicago:
I've been testing this out on the teststand and I am seeing some weird behavior:

1. Occasionally when I click "Overwrite Standard Run" it does not seem to update the DB values, and since they are no longer fetched from the database, my new settings are lost when I hit new run.

2. I got this error when trying to start a run just after saving a new version of a Maintenance run:

```
17-03-02 09:08:17 Error retrieving Standard Runs information: 
 -[NSNull unsignedLongValue]: unrecognized selector sent to instance 0xa0dbd140 
```

To reproduce the first issue on the teststand, just change whether PulseGT is enabled in the trigger mask, hit overwrite standard run, then uncheck it, save standard run, etc.

jcaravaca:
It should be properly refreshing now after saving. Tested at TS 3/7.

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -2006,19 +2122,24 @@ -(BOOL) loadStandardRun:(NSString*)runTypeName withVersion:(NSString*)runVersion
 }
 
 //Save MTC settings in a Standard Run table in CouchDB for later use by the Run Scripts or the user
--(BOOL) saveStandardRun:(NSString*)runTypeName withVersion:(NSString*)runVersion
+-(BOOL) saveStandardRun:(NSString*)runTypeName withVersion:(NSString*)runVersion;
```
tlatorre-uchicago: Typo? There is a semicolon at the end of the method definition.

jcaravaca: Good catch. Weird that XCodes doesn't complain... I'll fix that

tlatorre-uchicago:
merged.

PR #333, opened 2017-03-02
SNO+: fix caen event size calculation

tlatorre-uchicago:
The event size calculation in the caen controller was off by a factor of 2.


tlatorre-uchicago:
Tested on the teststand on March 2, 2017.

tlatorre-uchicago:
merged.

Issue #294, opened 2017-02-01
ping crates needs to check whether pedestal mask is set


tlatorre-uchicago:
We've seen a few cases where multiple crates fire at the same time, and when we had the debugging code on crate 1 it would fire for several events.

This can be done once #255 is merged.

tlatorre-uchicago:
Fixed in #343.

PR #343, opened 2017-03-10
SNO+: update ping crates

tlatorre-uchicago:
This commit updates the ping crates method to clear the pedestal mask for each
crate after firing pedestals. After firing pedestals on each crate, the
pedestal mask for each crate is set back to what it was before starting.

Fixes issue #294.


tlatorre-uchicago:
Tested on the teststand on March 13, 2017. However, since there is only one card on the teststand and the ECAL is not good, it was not a very definitive test. We might just have to wait and test this out on the detector later today.

tlatorre-uchicago:
merged.

PR #347, opened 2017-03-13
SNO+: update panic down method to launch a thread to send the command asynchronously

### LINKED ISSUE #346: Panic down all sends commands to crates serially

tlatorre-uchicago:
This should fix #346. I haven't tested this or even tried to compile it so it should be tested on the teststand. Ideally, since this will affect many crates, it should also be tested out on the detector with more than one crate ramped too.


tlatorre-uchicago:
Tested on the teststand on March 13, 2017. Since this affects multiple crates, it should be tested on the detector as well.

tlatorre-uchicago:
merged.

PR #348, opened 2017-03-13
update MTC model to load trigger scans after document is loaded

### LINKED ISSUE #318: mtc thresholds change after closing and reopening the window
I think this is because the thresholds are grabbed from the database in awakeFromNib. That should be done in a method called just once after the document loads, although they are already being set by the detector state database on bootup.

tlatorre-uchicago:
This pull request updates the MTC model to load the trigger scans after the document is loaded. It also contains a lot of formatting changes to make the code style in the MTC model and controller consistent. This fixes #318.


tlatorre-uchicago:
Tested at the teststand on March 13, 2017.

tlatorre-uchicago:
merged.

PR #349, opened 2017-03-13
open last standard run

### LINKED ISSUE #326: Orca should open up last standard run
When handing over the shift from u/g to a/g the operator has to remember to select the correct standard run before starting a new run. The last standard run should be saved to the .Orca file and selected when the standard runs are loaded from the database.

tlatorre-uchicago:
This pull request updates the SNO+ model to load the last standard run type and version saved to the Orca file when Orca boots up. This fixes #326.

This hasn't been tested yet.

### DIFF
--- a/Source/Experiments/SNOP/SNOPModel.m
+++ b/Source/Experiments/SNOP/SNOPModel.m
@@ -100,8 +100,6 @@ resync;
 
 #pragma mark ¥¥¥Initialization
 
-/* This method is obsolete and rarely called.
- For initialization you should use initWithCoder */
 - (id) init
 {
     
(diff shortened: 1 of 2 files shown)


Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -2016,42 +2017,34 @@ -(BOOL) refreshStandardRunsFromDB
...
-    //Check if DB is empty
-    if([[standardRunCollection objectForKey:[self standardRunType]] count] == 0){
+    if ([[standardRunCollection objectForKey:[self standardRunType]] count] == 0){
+        /* Database is empty, so we set the standard run version to the empty string. */
         [self setStandardRunVersion:@""];
-    }
-    //Check if previous selected run exists
-    else if([standardRunCollection objectForKey:[self standardRunType]] == nil){
-        //If not, select first on the list
+    } else if ([[standardRunCollection objectForKey:[self standardRunType]] objectForKey:[self standardRunVersion]] == nil) {
```
tlatorre-uchicago: @jcaravaca does this look right? Previously you were checking whether the standard run exists, but that is already guaranteed due to a check above. Here, I suspect you wanted to check if the version existed, so I updated it.

tlatorre-uchicago:
I updated the pull request to use the last selected standard run type and version instead of the last standard run which was run because the last standard run which was run is not saved to the Orca file when a new run starts. In addition, all of the settings in the Orca file load the last GUI state, so that is what is done here too.

I also fixed one bug in the standard run code and updated some of the formatting to make it consistent across the SNO+ model.

tlatorre-uchicago:
Tested on the teststand on March 13, 2017.

Source/Experiments/SNOP/SNOPModel.m
jcaravaca: Yes, this looks OK to me. If there is a SR type there is always a version so that's probably why I didn't catch this...

jcaravaca:
This feature was in the first SR version but I had to take it out due to having loads of log errors when starting ORCA if there was some problem. I think the SR are implemented in a cleaner way so that might not happen but we should test it thoroughly at the TS for every possible situation.

tlatorre-uchicago:
merged.

PR #351, opened 2017-03-13
SNO+: Only ramp down of supply A turns off crate triggers.

### LINKED ISSUE #344: Ramping down OWL supply causes Crate 16 triggers to be turned off

BenLand100:
Previously supply 16B would cause crate 16 triggers to turn off. Turning
off triggers for OWL PMTs on crate 16B ramp down not implemented yet
(unchanged from previous). Fixes #344.


tlatorre-uchicago:
merged.

PR #354, opened 2017-03-13
SNO+: update CAEN controller to load settings even if the user hasn't pressed enter

### LINKED ISSUE #353: Need to click enter when changing CAEN settings
It should be possible to load the values immediately when you click *Init Board*.

tlatorre-uchicago:
Fixes #353. Tested at the teststand on March 13, 2017 .


tlatorre-uchicago:
merged.

PR #358, opened 2017-03-14
TELLIE: Allow run files to be loaded through SNOPController

EdLeming:
This PR allows operators to use the SNOPController to request standardised TELLIE flash sequences down multiple fibres. It also updates the ELLIE gui to allow users to produce said standard run types, and push them to couch.

### DIFF
--- a/Source/Experiments/SNOP/SNOPController.m
+++ b/Source/Experiments/SNOP/SNOPController.m
@@ -65,8 +65,10 @@ const int view_mask_map[10] = {2,1,0,3,4,7,6,5,9,8};
 @synthesize
 tellieStandardSequenceFlag,
 tellieFireSettings,
-smellieRunFileList,
+smellieRunFileList = _smellieRunFileList,
+tellieRunFileList = _tellieRunFileList,
 smellieRunFile,
+tellieRunFile,
 snopBlueColor,
 snopRedColor,
 snopOrangeColor,
@@ -86,7 +88,6 @@ snopGreenColor;
 }
 - (void) dealloc
 {
-    [smellieRunFileList release];
     [smellieRunFile release];
     [snopBlueColor release];
     [snopGreenColor release];
@@ -415,10 +416,15 @@ snopGreenColor;
                         object: nil];
     
     [notifyCenter addObserver : self
-                     selector : @selector(fetchRunFilesFinish:)
+                     selector : @selector(fetchSmellieRunFilesFinish:)
                          name : @"SmellieRunFilesLoaded"
                         object: nil];
     
+    [notifyCenter addObserver : self
+                     selector : @selector(fetchTellieRunFilesFinish:)
+                         name : @"TellieRunFilesLoaded"
+                        object: nil];
+
 }
 
 - (void) updateWindow
@@ -537,10 +543,8 @@ snopGreenColor;
 
 - (IBAction) stopRunAction:(id)sender
 {
-
     [self endEditing];
     [model stopRun];
-
 }
 
 - (void) runStatusChanged:(NSNotification*)aNotification
@@ -964,7 +968,7 @@ snopGreenColor;
 //smellie functions ----------------------------------------------
 
 //this fetches the smellie run file information
-- (IBAction) fetchRunFiles:(id)sender
+- (IBAction) fetchSmellieRunFiles:(id)sender
 {
     // Temporarily disable drop down list and remove old items
     [smellieRunFileNameField addItemWithObjectValue:@""];
@@ -972,8 +976,6 @@ snopGreenColor;
     [smellieRunFileNameField setEnabled:NO];
     [smellieRunFileNameField removeAllItems];
     [smellieStartRunButton setEnabled:NO];
-    [smellieStopRunButton setEnabled:NO];
-    [smellieEmergencyStop setEnabled:NO];
     
     // Set the smellieRunFileList to nil
     [self setSmellieRunFileList:nil];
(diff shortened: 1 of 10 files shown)


Source/Experiments/SNOP/SNOPController.m
```diff
@@ -1017,93 +1057,93 @@ -(IBAction)loadSmellieRunAction:(id)sender
...
             if( [thisRunFile isEqualToString:requestedRunFile]){
                 
+                NSArray*  objs = [[(ORAppDelegate*)[NSApp delegate] document] collectObjectsOfClass:NSClassFromString(@"ELLIEModel")];
+                if (![objs count]) {
+                    NSString* reasonStr = @"ELLIE model not available, add an ELLIE model to your experiment";
+                    NSException* e = [NSException
+                                      exceptionWithName:@"NoEllieModel"
+                                      reason:reasonStr
+                                      userInfo:nil];
+                    [e raise];
```
tlatorre-uchicago: Is this going to be caught anywhere? If not, the exception will cause Orca to crash and no one will see the error message.

Source/Experiments/SNOP/SNOPController.m
```diff
@@ -1209,17 +1256,119 @@ - (IBAction) emergencySmellieStopAction:(id)sender
...
+                int no_nodes = [nodes count];
+                int rate = [[tellieRunFile objectForKey:@"trigger_rate"] intValue];
+                float total_time = (no_nodes*no_pulses)/(rate*60)*1.1;
+                [loadedTellieRunTimeLabel setStringValue:[NSString stringWithFormat:@"%0.1f",total_time]];
+            }
+        }
+        //Activate run buttons
+        [tellieStartRunButton setEnabled:YES];
+        [tellieStopRunButton setEnabled:NO];
+        //[tellieEmergencyStop setEnabled:NO];
```
tlatorre-uchicago: delete?

Source/Experiments/SNOP/SNOPController.m
```diff
@@ -1209,17 +1256,119 @@ - (IBAction) emergencySmellieStopAction:(id)sender
...
+    // Get ellie model and launch a fire
+    // sequence
+    NSArray*  objs = [[(ORAppDelegate*)[NSApp delegate] document] collectObjectsOfClass:NSClassFromString(@"ELLIEModel")];
+    if (![objs count]) {
+        NSString* reasonStr = @"ELLIE model not available, add an ELLIE model to your experiment";
+        NSException* e = [NSException
+                          exceptionWithName:@"NoEllieModel"
+                          reason:reasonStr
+                          userInfo:nil];
+        [e raise];
```
tlatorre-uchicago: same as above. Is this caught anywhere?

tlatorre-uchicago:
It seems like a lot of functionality is in the SNO+ controller. In general, the controller should really only be used to call model functions and do GUI updates. One problem for example, is that exceptions raised in the controller won't be caught anywhere and will cause Orca to crash.

Is it possible to move all of the heavy duty lifting to model functions?

Source/Experiments/SNOP/SNOPController.m
EdLeming: I have try catches around all calls to server functions. I was expecting the servers to throw exceptions when something is awry. This now just mimics the exception I was expecting the client to throw.

EdLeming:
Hmmm, yeah. This is one of those snowball moments, it was only ever meant to be a couple of lines in the controller, but things grow as the functionality evolves. I'll re-arrange (likely tomorrow) and push up.

Source/Experiments/SNOP/SNOPController.m
EdLeming: I'm waiting for the TUBii PR to go through to give this stop button a back end. I can delete for now.

EdLeming:
@tlatorre-uchicago I've moved the timing calculations out of the snopcontroller and into the ELLIEModel. The SNOPController now just:

1) Updates gui elements.
2) Performs some key checks on current run conditions.
3) Calls SNOP/ELLIE model functions. 
4) Hosts the TELLIE / SMELLIE run sequence threads

On the 4th point. We could move this to the SNOPModel, but I feel it would make the code slightly more opaque. Do you have a preference?

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -1758,6 +1764,45 @@ -(void) parseSmellieRunFileDocs:(id)aResult
     [[NSNotificationCenter defaultCenter] postNotificationName: @"SmellieRunFilesLoaded" object:nil];
 }
 
+- (void) getTellieRunFiles
+{
+    //Set SmellieRunFiles to nil
```
tlatorre-uchicago: Should be Tellie?

tlatorre-uchicago:
Yes, they should be moved to a model, becaues there can be multiple controllers for a single model when you duplicate a window. I'm not sure whether the SNOPModel or the ELLIE models are the better place to put them, but whatever makes things easier.

EdLeming:
@tlatorre-uchicago, I've moved the thread to the ELLIEModel. Local testing has been a success.

tlatorre-uchicago:
ok, if you can just squash to a single commit with a messge like:

```
SNO+: [main message]

[details]
```

I'll merge.

EdLeming:
👍  I'll get this pushed up in the morning

EdLeming:
@tlatorre-uchicago. This guy should be ready to go.

tlatorre-uchicago:
Could you remove the empty whitespace from the commit?

tlatorre-uchicago:
merged.

PR #359, opened 2017-03-15
Added keep alive thread to TUBii model

EdLeming:
This PR is to provide a similar emergency stop for TELLIE as we have for SMELLIE via the keep alive ORCA sends to the lasers interlock server. The idea being, if TUBii misses a keep alive (possibly due to ORCA dying), it stops sending triggers to the ELLIE systems. 

I spoke with @icoulter at the collab meeting about this and I believe he's already implemented the back end, although I'll leave it to him to confirm. One thing I don't know is the function on the TUBii server ORCA should be calling. As a place holder I've used "keepAlive", but I expect to have to update it.

### DIFF
--- a/Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiModel.m
+++ b/Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiModel.m
@@ -30,6 +30,8 @@ NSString* ORTubiiLock				= @"ORTubiiLock";
 @implementation TUBiiModel
 #pragma mark •••Synthesized Variables
 
+@synthesize keepAliveThread = _keepAliveThread;
+
 - (void) setUpImage
 {
     NSImage* img = [NSImage imageNamed:@"tubii"];
@@ -147,14 +149,27 @@ NSString* ORTubiiLock				= @"ORTubiiLock";
     [self ResetFifo]; //Maybe take this out eventually? I'm not sure
     [self setDataReadout:YES];
 }
+
+- (void) awakeAfterDocumentLoaded
+{
+    [self activateKeepAlive];
+}
+
 #pragma mark •••Network Communication
 - (void) sendOkCmd:(NSString* const)aCmd {
+    [self sendOkCmd:aCmd print:YES];
+}
+- (void) sendOkCmd:(NSString* const)aCmd print:(BOOL)printCheck{
     @try {
-        NSLog(@"Sending %@ to TUBii\n",aCmd);
+        if(printCheck){
+            NSLog(@"Sending %@ to TUBii\n",aCmd);
+        }
         [connection okCommand: [aCmd UTF8String]];
     }
     @catch (NSException *exception) {
-        NSLogColor([NSColor redColor],@"Command: %@ failed.  Reason: %@\n", aCmd,[exception reason]);
+        if(printCheck){
+            NSLogColor([NSColor redColor],@"Command: %@ failed.  Reason: %@\n", aCmd,[exception reason]);
+        }
     }
 }
 - (int) sendIntCmd: (NSString* const) aCmd {
@@ -737,4 +752,59 @@ NSString* ORTubiiLock				= @"ORTubiiLock";
     aState.controlReg = [self controlReg];
     return aState;
 }
+
+//////////////////////////////////////////
+// Keep alive stuff -
+// If ORCA dies we want to force tubii from
+// sending triggers to the ellie system
+-(void)activateKeepAlive
+{
+    /*
+     Start a thread to constantly send a keep alive signal to the smellie interlock server
+     */
+    [self setKeepAliveThread:[[[NSThread alloc] initWithTarget:self selector:@selector(pulseKeepAlive:) object:nil] autorelease]];
+    [[self keepAliveThread] start];
+}
+
+-(void)pulseKeepAlive:(id)passed
+{
+    /*
+     A fuction to be run in a thread, continually sending keep alive pulses to the interlock server
+     */
+    NSAutoreleasePool *pool = [[NSAutoreleasePool alloc] init];
+    
+    int counter = 0;
+    while (![[self keepAliveThread] isCancelled]) {
+        @try{
+            dispatch_sync(dispatch_get_main_queue(), ^{
+                [self sendOkCmd:@"keepAlive" print:NO];
+            });
+        } @catch(NSException* e) {
+            NSLogColor([NSColor redColor], @"[TUBii]: Problem sending keep alive to TUBii server, reason: %@\n", [e reason]);
+            break;
+        }
+        [NSThread sleepForTimeInterval:0.5];
+
+        // This is a very long running thread need to relase the pool every so often
+        if(counter == 1000){
+            [pool release];
+            pool = [[NSAutoreleasePool alloc] init];
+            counter = 0;
+        }
+        counter = counter + 1;
+    }
+    NSLog(@"[TUBii]: Stopped sending keep-alive to TUBii - ELLIE pulses will be shut off\n");
+    [pool release];
+}
+
+-(void)killKeepAlive
+{
+    /*
+     Stop pulsing the keep alive and disarm the interlock
+     */
+    [[self keepAliveThread] cancel];
+    NSLog(@"[TUBii]: Killing keep alive - ELLIE pulses will be shut off\n");
+}
+
+
 @end
--- a/Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiModel.h
+++ b/Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiModel.h
@@ -109,6 +109,7 @@ struct TUBiiState { //A struct that allows users of TUBiiModel to get/set all of
     RedisClient *connection;
     int portNumber;
     NSString* strHostName;//"192.168.80.25";
+    NSThread* _keepAliveThread;
 }
 @property (readonly) BOOL solitaryObject; //Prevents there from being two TUBiis
 @property (nonatomic) int portNumber;
@@ -134,6 +135,7 @@ struct TUBiiState { //A struct that allows users of TUBiiModel to get/set all of
 @property (nonatomic) BOOL TUBiiIsLOSrc;
 @property (nonatomic) BOOL CounterMode;
 @property (nonatomic) struct TUBiiState CurrentState; //Get/Sets a struct that fully specifies TUBii's Current state
+@property (nonatomic,retain) NSThread* keepAliveThread;
 
 #pragma mark •••Initialization
 - (id) init;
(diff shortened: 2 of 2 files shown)


icoulter:
Hi Ed,
The relevant TUBii code is here:
https://github.com/icoulter/workspace_tubii/commit/369dc68872280b9f25338e96d5e68283906da232
It hasn't been tested yet as I think we needed both ends of this. I think I can test it but probably not until I get back to Penn next week.
How often does your thread send the KeepAlive command? TUBii looks for it every second but as discussed we can tune this. Might be good to have something like a 1:2 ratio of rates just to ensure against false alarms (but we'll test this).

I've also implemented a TUBii "Stop" command which can stop TUBii's pulsers and trigger mask manually. Thinking that'd be useful to add to the gui but I can do that at a later date (probably next week) if you prefer.

Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiModel.m
```diff
@@ -737,4 +738,50 @@ - (struct TUBiiState) CurrentState {
...
+//////////////////////////////////////////
+// Keep alive stuff -
+// If ORCA dies we want to force tubii from
+// sending triggers to the ellie system
+-(void)activateKeepAlive
+{
+    /*
+     Start a thread to constantly send a keep alive signal to the smellie interlock server
+     */
+    NSAutoreleasePool *pool = [[NSAutoreleasePool alloc] init];
```
tlatorre-uchicago: Is this run in a separate thread? If not, I think you want the pool in the other thread and not here.

EdLeming:
Hey @icoulter. Cool, thanks - I was right with keepAlive! At the moment it's programmed to send the keep alive request at 10Hz. It will likely be a little slower than that due to network issues, but it should still give us plenty of breathing room. 

I can see how it would be useful to have in the GUI, but it's more useful to have the functionality called when a user  hits the "Stop (T/SM)ELLIE" buttons in the SNOPController (that way it's only one button push to do it all).  Currently I call the stop(T/SM)elliePulser commands here in those button pushes. However, I'm not sure if they also affect the trigger mask?

Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiModel.m
```diff
@@ -737,4 +738,50 @@ - (struct TUBiiState) CurrentState {
...
+-(void)pulseKeepAlive:(id)passed
+{
+    /*
+     A fuction to be run in a thread, continually sending keep alive pulses to the interlock server
+     */
+    NSAutoreleasePool *pool = [[NSAutoreleasePool alloc] init];
+
+    while (![[self keepAliveThread] isCancelled]) {
+        @try{
+            [self sendOkCmd:@"keepAlive"];
```
tlatorre-uchicago: The RedisClient class is currently not thread safe, so this needs to be sent to the main thread.

EdLeming:
So, to address @tlatorre-uchicago 's comment I now launch a thread, but I get it post the redisClient requests on the main thread. 

In an attempt to avoid slow downs on gui clicks (since they're also on the main thread) I've slowed to 2 Hz. We might consider reducing this further. In the event of a crash, killing ELLIE on the second scale seems reasonable. We're really just trying to stop the case where we get a run away flash sequence that would run for minutes out of our control.

Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiModel.m
```diff
@@ -737,4 +740,52 @@ - (struct TUBiiState) CurrentState {
...
+    NSLog(@"[TUBii]: Stopped sending keep-alive to TUBii - ELLIE pulses will be shut off\n");
+    [pool release];
+}
+
+-(void)killKeepAlive
+{
+    /*
+     Stop pulsing the keep alive and disarm the interlock
+     */
+    NSAutoreleasePool *pool = [[NSAutoreleasePool alloc] init];
```
tlatorre-uchicago: Where is this called from? If it's the main thread, you don't need to create an auto release pool here.

icoulter:
Testing this at the minute.
Its pretty noisy on the orca log. It prints a statement everytime it sends the keepAlive command.

icoulter:
Tested this locally and it seems to work with the latest TUBii firmware.
We'll have to quieten it down a bit though. I can take a look at that.

EdLeming:
Great news! Thanks for testing it out. I guess the noisy-ness is coming from -(void) sendOkCmd?

icoulter:
Yeah. I don't know if there's a silent equivalent but I can try to find out.

EdLeming:
We could just pass a flag into that function. I'll get it done when I have a second. 👍

EdLeming:
@icoulter, try that.

icoulter:
That works. Its also similarly noisy with connection failed outputs if it can't reach TUBii so that might be worth quietening down as well.

icoulter:
That works 🎉

tlatorre-uchicago:
Is this ready to go? If so, can you squash it down to a single commit?

tlatorre-uchicago:
@icoulter, just to confirm this is already implemented and deployed on the TUBii side?

Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiModel.m
```diff
@@ -737,4 +747,50 @@ - (struct TUBiiState) CurrentState {
...
+//////////////////////////////////////////
+// Keep alive stuff -
+// If ORCA dies we want to force tubii from
+// sending triggers to the ellie system
+-(void)activateKeepAlive
+{
+    /*
+     Start a thread to constantly send a keep alive signal to the smellie interlock server
+     */
+    [self setKeepAliveThread:[[NSThread alloc] initWithTarget:self selector:@selector(pulseKeepAlive:) object:nil]];
```
tlatorre-uchicago: need to free the thread if you alloc it. maybe better to call autorelease?

icoulter:
Implemented and committed to git but not deployed yet.

Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiModel.m
```diff
@@ -737,4 +747,50 @@ - (struct TUBiiState) CurrentState {
...
+    [[self keepAliveThread] start];
+}
+
+-(void)pulseKeepAlive:(id)passed
+{
+    /*
+     A fuction to be run in a thread, continually sending keep alive pulses to the interlock server
+     */
+    NSAutoreleasePool *pool = [[NSAutoreleasePool alloc] init];
```
tlatorre-uchicago: I just realized that if this is going to be a very long running thread, you should figure out some way to release the pool every so often otherwise it will never get released. You can either just move to releasing it and creating it within the loop, or create some counter that releases it every few iterations of the loop.

tlatorre-uchicago:
@icoulter, ok let me know when it is deployed.

icoulter:
Ok, I may have to do it tomorrow as I'm at home waiting on a delivery and don't have access to the VPN on my home laptop. Unless if you want to do it today, I can talk someone else through it.

EdLeming:
@icoulter. Are you happy with how this performed underground the other day? If so I'll address @tlatorre-uchicago's comment about the autorealse pool and this should be good to go.

icoulter:
Yes. Sorry, I forgot to send you that email. I think I'm happy that it works as intended.

Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiModel.m
```diff
@@ -737,4 +747,58 @@ - (struct TUBiiState) CurrentState {
...
+                [self sendOkCmd:@"keepAlive" print:NO];
+            });
+        } @catch(NSException* e) {
+            NSLogColor([NSColor redColor], @"[TUBii]: Problem sending keep alive to TUBii server, reason: %@\n", [e reason]);
+            return;
+        }
+        [NSThread sleepForTimeInterval:0.5];
+
+        // This is a very long running thread need to relase the pool every so often
+        if(counter == 1000){
```
tlatorre-uchicago: Where does the counter get incremented?

Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiModel.m
```diff
@@ -737,4 +747,58 @@ - (struct TUBiiState) CurrentState {
...
+    
+    int counter = 0;
+    while (![[self keepAliveThread] isCancelled]) {
+        @try{
+            dispatch_sync(dispatch_get_main_queue(), ^{
+                [self sendOkCmd:@"keepAlive" print:NO];
+            });
+        } @catch(NSException* e) {
+            NSLogColor([NSColor redColor], @"[TUBii]: Problem sending keep alive to TUBii server, reason: %@\n", [e reason]);
+            return;
```
tlatorre-uchicago: need to release the pool before here.

tlatorre-uchicago: actually, probably better to just break so that it logs that the keep alive is not being sent anymore.

Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiModel.m
```diff
@@ -118,6 +120,7 @@ - (id) initWithCoder:(NSCoder *)aCoder {
             [self setPortNumber:TUBII_DEFAULT_PORT];
         }
     }
+    [self activateKeepAlive];
```
tlatorre-uchicago: This should be called after the document loads. To do that you can create a function in the TUBii model called `awakeAfterDocumentLoaded` which will automatically get called when the document gets loaded. See https://github.com/snoplus/orca/blob/master/Source/Objects/Custom%20Hardware/SNO%2B/MTC/ORMTCModel.m#L152 for an example.

tlatorre-uchicago:
Last thing is to either free or autorelease the thread and then I think this is good.

Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiModel.m
```diff
@@ -737,4 +752,60 @@ - (struct TUBiiState) CurrentState {
...
+        if(counter == 1000){
+            [pool release];
+            pool = [[NSAutoreleasePool alloc] init];
+            counter = 0;
+        }
+        counter = counter + 1;
+    }
+    NSLog(@"[TUBii]: Stopped sending keep-alive to TUBii - ELLIE pulses will be shut off\n");
+    [pool release];
+    [[self keepAliveThread] release];
```
tlatorre-uchicago: Not sure if it's kosher to release the thread that you are currently in. I would just call autorelease on the thread when you create it. Then, when you set it (since you declared the property as retain) it will get retained which will make sure it's not freed until you set another one at which point it should be released.

tlatorre-uchicago:
merged.

PR #361, opened 2017-03-16
SMELLIE: Properly handle exceptions passed back by smellie server

EdLeming:
This PR addresses the issue that exceptions on the SMELLIE server were not being properly caught by ORCA. The server actually catches its own (internally thrown) exceptions and wraps them into a sting before passing back to ORCA. A standard try / catch in ORCA is therefore not effective.  

Two other minor issues that came up in testing are also addressed: 

Non-expert ELLIE daq users had tried to use the 'standard' stop run button in the SNOPController to stop the ELLIE system firing. I've added a check in that button push to make sure that, if required, the appropriate ELLIE stop methods are called before the run stops. 

Extended the timeout on the TELLIE server. Since moving the tellie server to SNOROP communication between the server and the TELLIE hardware is taking significantly longer than has been observed with Mac or Ubuntu systems. The ELLIE group are working on speed ups, but a extended timeout is necessary in the  short term.

### DIFF
--- a/Source/Experiments/SNOP/SNOPController.h
+++ b/Source/Experiments/SNOP/SNOPController.h
@@ -160,7 +160,6 @@
     // Detector State
     IBOutlet WebView* detectorState;
 }
-
 @property (nonatomic) BOOL tellieStandardSequenceFlag;
 @property (nonatomic,retain) NSDictionary *tellieFireSettings;
 @property (nonatomic,retain) NSMutableDictionary *tellieRunFileList;
@@ -242,3 +241,4 @@
 @end
 
 extern NSString* ORSNOPRequestHVStatus;
+extern NSString* ORRunWaitFinished;
--- a/Source/Experiments/SNOP/SNOPController.m
+++ b/Source/Experiments/SNOP/SNOPController.m
@@ -38,6 +38,8 @@
 #import "ECARun.h"
 
 NSString* ORSNOPRequestHVStatus = @"ORSNOPRequestHVStatus";
+NSString* ORRunWaitFinished = @"ORRunWaitFinished";
+
 
 #define UNITS_UNDECIDED 0
 #define UNITS_RAW       1
@@ -424,7 +426,6 @@ snopGreenColor;
                      selector : @selector(fetchTellieRunFilesFinish:)
                          name : @"TellieRunFilesLoaded"
                         object: nil];
-
 }
 
 - (void) updateWindow
@@ -544,7 +545,25 @@ snopGreenColor;
 - (IBAction) stopRunAction:(id)sender
 {
     [self endEditing];
-    [model stopRun];
+
+    NSArray*  objs = [[(ORAppDelegate*)[NSApp delegate] document] collectObjectsOfClass:NSClassFromString(@"ELLIEModel")];
+    if (![objs count]) {
+        NSLogColor([NSColor redColor], @"ELLIE model not available, add an ELLIE model to your experiment\n");
+        goto err; // If error fall through and just stop the run.
+    }
+    ELLIEModel* theELLIEModel = [objs objectAtIndex:0];
+
+    if([[theELLIEModel tellieThread] isExecuting]){
+        [theELLIEModel stopTellieRun];
+    }
+    if([[theELLIEModel smellieThread] isExecuting]){
+        [theELLIEModel stopSmellieRun];
+    }
+
+err:
+    {
+        [model stopRun];
+    }
 }
 
 - (void) runStatusChanged:(NSNotification*)aNotification
@@ -1167,16 +1186,6 @@ snopGreenColor;
     }
     ELLIEModel* theELLIEModel = [objs objectAtIndex:0];
 
-    //////////////////////
-    // Check if the thread is still running,
-    // if not cancel the thread and let the
-    // catch statements in [ellieModel startSmellieRun]
-    // cause a goto err;
-    if(![[theELLIEModel smellieThread] isCancelled]){
-        [[theELLIEModel smellieThread] cancel];
-        return;
-    }
-
     //Call stop smellie run method to tidy up SMELLIE's hardware state
     @try{
         [theELLIEModel stopSmellieRun];
@@ -1188,9 +1197,7 @@ snopGreenColor;
     ////////////
     // Roll over into maintinance run
     if([[model lastStandardRunType] isEqualToString:@"SMELLIE"]){
-        [model setStandardRunType:@"MAINTENANCE"];
-        [self loadStandardRunFromDBAction:self];
-        [self startRunAction:self];
+        [model startStandardRun:@"MAINTENANCE" withVersion:@"DEFAULT"];
     }
 }
 
@@ -1379,16 +1386,6 @@ snopGreenColor;
     }
     ELLIEModel* theELLIEModel = [objs objectAtIndex:0];
 
-    //////////////////////
-    // Check if the thread has been cancelled,
-    // if not cancel the thread and let the
-    // catch statements in [ellieModel startTellieRun]
-    // cause a goto err;
-    if(![[theELLIEModel tellieThread] isCancelled]){
-        [[theELLIEModel tellieThread] cancel];
-        return;
-    }
-    
     //Call stop smellie run method to tidy up TELLIE's hardware state
     @try{
         [theELLIEModel stopTellieRun];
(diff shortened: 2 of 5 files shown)


EdLeming:
@jcaravaca it might be worth having a scan over this. Jeff mentioned it spotted some crashes on changing between standard runs, although it's not clear if they're associated with these changes. 

Your expert eye would be useful to make sure I'm not interfering with anything.

tlatorre-uchicago:
For the run stop, is it possible to just check if the smellie or tellie thread is executing instead of having another flag which may get out of sync?

jcaravaca:
@EdLeming, sure, I'll have a look shortly

Source/Experiments/SNOP/SNOPController.m
```diff
@@ -1190,6 +1195,10 @@ - (IBAction) stopSmellieRunAction:(id)sender
     }
 
     ////////////
+    // Set run flag
+    [self setSmellieRunFlag:NO];
+
+    ////////////
     // Roll over into maintinance run
     if([[model lastStandardRunType] isEqualToString:@"SMELLIE"]){
         [model setStandardRunType:@"MAINTENANCE"];
```
jcaravaca: You should substitute this block by [aSNOPModel startStandardRun:@"MAINTENANCE" withVersion:@"DEFAULT"]; instead of using the actions

EdLeming: 👍

jcaravaca:
@EdLeming the crashes we were experiencing with Jeff during the the testing of this branch are due to a problem with the standard run GUI and nothing to do with this. I just created a PR that hopefully fixes it.

Source/Experiments/SNOP/SNOPController.m
```diff
@@ -534,10 +534,14 @@ - (IBAction)resyncRunAction:(id)sender
 
 - (IBAction) stopRunAction:(id)sender
 {
-
+    if([smellieThread isExecuting]){
+        [self stopSmellieRunAction:nil];
```
tlatorre-uchicago: Does this actually wait for the thread to be cancelled? If not, I think it needs to do some more complicated gymnastics here to wait for the thread to stop. The problem is when someone hits stop run, this may cancel the thread but it will still be injecting light for a while. Since the run sequence doesn't know this it will roll into the next run and we will have light in that run.

I think the right solution is that whatever function stops the run should:

1. cancel the thread
2. post a notification telling the run control to wait
3. launch a thread that waits for the thread to stop
4. new thread checks if tellie/smellie thread is stopped and when it does it posts a notification telling the run to continue

EdLeming:
@tlatorre-uchicago. Have a look at the most recent commit. I can't test locally as I can't actually start a run, but I think it should do the trick.

tlatorre-uchicago:
This won't actually cause the run to wait. I was imagining something like:

In SMELLIEModel:

```objective-c
- (void) isRunning
{
    return [smellieThread isExecuting];
}

- (void) stopRun
{
    [smellieThread cancel];

    /* Post a notification telling the run control to wait until the thread finishes. */
    NSDictionary *userInfo = [NSDictionary dictionaryWithObjectsAndKeys:
                                  @"waiting for SMELLIE run to finish", @"Reason",
                                  nil];
        [[NSNotificationCenter defaultCenter] postNotificationName:ORAddRunStateChangeWait object: self userInfo: userInfo];

        /* detach a thread to monitor XL3/CAEN/MTC buffers */
        [NSThread detachNewThreadSelector:@selector(_waitForSmellieRunToFinish)
                                 toTarget:self
                               withObject:nil];
}

- (void) waitForSmellieRunToFinish
{
    @autoreleasepool {
        while ([smellieThread isExecuting]) {
        }

        /* Go ahead and end the run. */
        dispatch_sync(dispatch_get_main_queue(), ^{
            [[NSNotificationCenter defaultCenter] postNotificationName:ORReleaseRunStateChangeWait object: self];
        });
    }
}
```

EdLeming:
@tlatorre-uchicago. I see. I forgot the return statements after my thread stops. If I put them in it should do the same thing, no?

tlatorre-uchicago:
Where do you post the notification telling the run to not stop? I don't really understand the run stop thread right now, it seems to not be doing anything, i.e. it gets launched when you press run stop, but the function continues and stops the run anyways. Am I missing something?

Posting a notification to hold off the run stop is not necessary if you are guaranteed that tellie/smellie will not be firing by the time `stopSmellieRunAction` and `stopTellieRunAction` complete, but it wasn't obvious if those methods just ask the thread to quit and that there may be some period of time after they return when they are still firing light (even a short period of time).

EdLeming:
hey @tlatorre-uchicago. I've added in the return statements and posted the final method call to the main thread. Have a look. I think those missing return statements were the source of the confusion.

tlatorre-uchicago:
OK, I understand now. I see a few potential issues with this solution though:

- If for whatever reason, the thread doesn't stop this will enter an infinite loop and the user will not have any recourse. If you post the run wait notification above, there is at least the option to both view what is currently holding the run from ending and to clear them if necessary. Currently these are only available in the Orca run control, but I think there is a plan to add them to the SNO+ control panel.
- I noticed that in the `stopSmellieRunAction` method of the SNO+ controller it will start a new maintenance run. If I understand the code correctly, when you hit stop run it is possible that it starts a new maintenance run and then immediately stops it when that thread finishes.

I think the `runStopThread` also needs to be freed somewhere.

EdLeming:
@tlatorre-uchicago. OK, I see. I hadn't realised the run model had that extra functionality. 

I moved the TELLIE / SMELLIE threads to be hosted in the ELLIEModel in #358, which I believe this is ready to get pulled through. It's probably sensible to wait for that to happen before I implement your suggested changes. The I'll get this up asap.

EdLeming:
@tlatorre-uchicago. I've updated the logic associated with ending both TELLIE and SMELLIE flash sequences. Now all threading and coordination (including posting the requested run wait notification) happens in the ELLIEModel. 

I can't start runs locally so have not been able to test the response to the ORAddRunStateChangeWait notifications. For example, I have not tested the roll over into maintenance, here:

https://github.com/EdLeming/orca/blob/string_exceptions/Source/Experiments/SNOP/SNOPController.m#L1384

If I understand correctly this request should be queued until ORReleaseRunStateChangeWait is posted. This happens at the end of the [theELLIEModel stopTellieRun] sequence.

tlatorre-uchicago:
Do you have any fake TELLIE servers running on the teststand? If so you could test the run starts from there.

EdLeming:
Ah, I don't currently, but I could. I'll try that this afternoon.

EdLeming:
@tlatorre-uchicago. Testing was a success. Have a look over when you have time and let me know what you think.

Source/Experiments/SNOP/SNOPController.h
```diff
@@ -113,6 +113,7 @@
     BOOL tellieStandardSequenceFlag;
     IBOutlet NSButton* runsLockButton;
     IBOutlet NSTextField *lockStatusTextField;
+    NSThread* _runStopThread; // A thread to check RUN STOP button hits coordinate with ELLIE runs
```
tlatorre-uchicago: Can this be deleted?

Source/Experiments/SNOP/SNOPController.h
```diff
@@ -173,6 +173,7 @@
 @property (nonatomic,retain) NSColor *snopOrangeColor;
 @property (nonatomic,retain) NSColor *snopBlackColor;
 @property (nonatomic,retain) NSColor *snopGrayColor;
+@property (nonatomic, retain) NSThread* runStopThread;
```
tlatorre-uchicago: Here too.

Source/Experiments/SNOP/SNOPController.m
```diff
@@ -421,10 +424,14 @@ - (void) registerNotificationObservers
                         object: nil];
     
     [notifyCenter addObserver : self
+                     selector : @selector(stopRunAction:)
```
tlatorre-uchicago: This can be deleted?

EdLeming:
A reasonable question. The logic is that, even if the thread errors, the flash sequence always gets ended as if a user had hit the stop button. This avoids behavioural inconsistencies between the two possible routes to stopping a sequence. 

That said, now we've moved most of the functionality associated with stopping a sequence to the ELLIEModel it's no longer such an obvious path to take. The one element left to the stop button which is not handled in the ELLIEModel is rolling over into an appropriate run when a sequence has ended. I quite like this being in the SNOPController as the button push is now a 'high level' control, coordinating the ELLIE specific behaviour (handled by ELLIEModel methods) and the run control. 

If you're happy enough with that I've tidied into a nice single commit.

tlatorre-uchicago:
merged.

PR #365, opened 2017-03-19
SNO: Try to solve SNO+ GUI crashes.

jcaravaca:

### DIFF
--- a/Source/Experiments/SNOP/SNOPController.m
+++ b/Source/Experiments/SNOP/SNOPController.m
@@ -481,9 +481,8 @@ snopGreenColor;
 - (IBAction) startRunAction:(id)sender
 {
     //Load selected SR in case the user didn't click enter
-    //fixed memory leak in the next two lines -- don't 'copy without a matching release MAH 03/8/2017
-    NSString *standardRun = [standardRunPopupMenu objectValueOfSelectedItem];
-    NSString *standardRunVersion = [standardRunVersionPopupMenu objectValueOfSelectedItem];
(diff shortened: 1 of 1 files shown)


jcaravaca:
This is an attempt to fix the SNO+ GUI crashes due to the standard runs selection.
This PR is against tag0.6 intended to be deployed on site. I'll rebase with master if the fix works and resubmit.

tlatorre-uchicago:
In all three functions updated, there are early return statements which will leak memory.

jcaravaca:
According to Mark Howe, early returns do not leak memory as long as the copies are matching a release statement.

tlatorre-uchicago:
Huh? I'm not sure I understand. What do you mean by "the copies are matching a release statement"?

jcaravaca:
Literally from Mark Howe:
> As long as the copies are paired with releases, even in the cases were an early return can happen that would not leak.

tlatorre-uchicago:
I'm assuming he meant that you won't leak memory as long as you release before the early return.

jcaravaca:
Hmm, that might make more sense. I was assuming some Coca magic here...

tlatorre-uchicago:
Are you convinced that this has fixed the problem? If so, I can merge this later today and we can make a new release.

jcaravaca:
Before the change it used to crash 1 out of 3 times we changed a run.
After the change on Saturday we haven't had any crash.

tlatorre-uchicago:
merged.

PR #367, opened 2017-03-19
SNO+: changed start/resync logic and some GUI improvements

jcaravaca:
- [x] Keep window unlocked when we switch through XL3s, FECs and DBs
- [x] Lock MTC when toggling radio buttons in non-maintenance
- [x] No polling in Physics
- [x] Ask to stop polling when starting a non-maintenance run
- [x] Add subrun number to run control display
- [x] Lock field in CAEN
- [x] Show trigger status in HV Master
- [x] Add OWLs HV state in HV master
- [x] Add tooltips for standard runs
- [x] Lock Triggers ON/OFF when in physics

### DIFF
--- a/Source/Experiments/SNOP/ECARun.m
+++ b/Source/Experiments/SNOP/ECARun.m
@@ -143,7 +143,7 @@ NSString* ORECARunFinishedNotification = @"ORECARunFinishedNotification";
         if ([objs count]) {
             anMTCModel = [objs objectAtIndex:0];
             [anMTCModel setPgtRate:[[self ECA_rate] floatValue]]; //UNCOMMENT THIS LINE BEFORE COMMISSIONING
-            [anMTCModel loadPulserRateToHardware];
+            [anMTCModel loadPulserRateToHardware]; //UNCOMMENT THIS LINE BEFORE COMMISSIONING
         }
     }
 
--- a/Source/Experiments/SNOP/SNOPController.m
+++ b/Source/Experiments/SNOP/SNOPController.m
@@ -1497,9 +1497,11 @@ err:
     
     //Enable or disable fields
     [standardRunThresCurrentValues setEnabled:!lockedOrNotRunningMaintenance];
-    [standardRunSaveButton setEnabled:!lockedOrNotRunningMaintenance];
-    [standardRunLoadButton setEnabled:!lockedOrNotRunningMaintenance];
+    [standardRunSaveButton setEnabled:!locked];
+    [standardRunLoadButton setEnabled:!locked];
     [standardRunLoadinHWButton setEnabled:!lockedOrNotRunningMaintenance];
+    [triggersOFFButton setEnabled:notRunningOrInMaintenance];
+
     //Do not lock detector state bits to the operator
     for(int irow=0;irow<21;irow++){
         [[runTypeWordMatrix cellAtRow:irow column:0] setEnabled:!lockedOrNotRunningMaintenance];
--- a/Source/Experiments/SNOP/SNOPController.m
+++ b/Source/Experiments/SNOP/SNOPController.m
@@ -1234,10 +1234,7 @@ err:
     ////////////
     // Roll over into maintenance run
     if([[model lastStandardRunType] isEqualToString:@"SMELLIE"]){
-        [model setStandardRunType:@"MAINTENANCE"];
-        [model loadStandardRun:[model standardRunType] withVersion:[model standardRunVersion]];
-        [model loadSettingsInHW];
-        [self startRunAction:self];
+        [model startStandardRun:@"MAINTENANCE" withVersion:@"DEFAULT"];
     }
 }
 
@@ -1439,10 +1436,7 @@ err:
     if([[model lastStandardRunType] isEqualToString:@"TELLIE"]){
         // If user was running a TELLIE standard sequence, roll over into maintinance run
         if([self tellieStandardSequenceFlag]){
-            [model setStandardRunType:@"MAINTENANCE"];
-            [model loadStandardRun:[model standardRunType] withVersion:[model standardRunVersion]];
-            [model loadSettingsInHW];
-            [self startRunAction:self];
+            [model startStandardRun:@"MAINTENANCE" withVersion:@"DEFAULT"];
         // If user is using the ellie gui simply start a new run as they'll likely need to run
         // more sequences. Reasonable as this is an 'expert' level operation. Proceedures
         // will dictate the user should start a new standard run manualy when they're finished
--- a/Source/Experiments/SNOP/SNOPController.m
+++ b/Source/Experiments/SNOP/SNOPController.m
@@ -496,14 +496,6 @@ snopGreenColor;
 - (IBAction) startRunAction:(id)sender
 {
 
-    /* Start/Restart forces a resync */
-
-    /* A resync run does a hard stop and start without the user having to hit
-     * stop run and then start run. Doing this resets the GTID, which resyncs
-     * crate 9 after it goes out of sync :). */
-
-    [model setResync:YES];
-
     //If we are in OPERATOR mode we don't allow other version than DEFAULT
     BOOL locked = [gSecurity isLocked:ORSNOPRunsLockNotification];
     if(locked) { //Operator Mode
@@ -531,6 +523,19 @@ snopGreenColor;
 
     //Start the standard run and stop run initialization if failed
     if(![model startStandardRun:[model standardRunType] withVersion:[model standardRunVersion]]) return;
+
+}
+
+- (IBAction)resyncRunAction:(id)sender
+{
+
+    /* A resync run does a hard stop and start without the user having to hit
+     * stop run and then start run. Doing this resets the GTID, which resyncs
+     * crate 9 after it goes out of sync :). */
+
+    [model setResync:YES];
+
+    [self startRunAction:nil];
 
 }
 
@@ -566,6 +571,7 @@ err:
         [startRunButton setTitle:@"RESTART"];
         [lightBoardView setState:kGoLight];
         [runStatusField setStringValue:@"Running"];
+        [resyncRunButton setEnabled:true];
         [runNumberField setStringValue:[runControl fullRunNumberString]];
         [doggy_icon start_animation];
 	}
@@ -573,6 +579,7 @@ err:
         [startRunButton setTitle:@"START"];
         [lightBoardView setState:kStoppedLight];
         [runStatusField setStringValue:@"Stopped"];
+        [resyncRunButton setEnabled:false];
         [doggy_icon stop_animation];
 	}
 	else if([runControl runningState] == eRunStarting || [runControl runningState] == eRunStopping || [runControl runningState] == eRunBetweenSubRuns){
@@ -580,6 +587,7 @@ err:
             //The run started so update the display
             [runStatusField setStringValue:@"Starting"];
             [startRunButton setEnabled:false];
+            [resyncRunButton setEnabled:false];
             [startRunButton setTitle:@"STARTING..."];
         }
         else {
@@ -1483,7 +1491,7 @@ err:
     BOOL runInProgress				= [gOrcaGlobals runInProgress];
     BOOL locked						= [gSecurity isLocked:ORSNOPRunsLockNotification];
     BOOL lockedOrNotRunningMaintenance = [gSecurity runInProgressButNotType:eMaintenanceRunType orIsLocked:ORSNOPRunsLockNotification];
-    BOOL notRunningOrInMaintenance = [model isNotRunningOrIsInMaintenance];
+    BOOL notRunningOrInMaintenance = isNotRunningOrIsInMaintenance();
 
     //[softwareTriggerButton setEnabled: !locked && !runInProgress];
     [runsLockButton setState: locked];
--- a/Source/Experiments/SNOP/SNOPController.m
+++ b/Source/Experiments/SNOP/SNOPController.m
@@ -521,15 +521,13 @@ snopGreenColor;
     if( !((dbruntypeword & kMaintenanceRun) || (dbruntypeword & kDiagnosticRun)) ){
         //Make sure we are not polling
         NSArray* xl3s = [[(ORAppDelegate*)[NSApp delegate] document] collectObjectsOfClass:NSClassFromString(@"ORXL3Model")];
-        bool xl3ispolling = false;
         for(ORXL3Model *anXL3 in xl3s){
-            if([anXL3 isPollingXl3]) xl3ispolling = true;
+            if([anXL3 isPollingXl3]) {
+                NSLog(@"Stopping XL3 polling on crate %d\n",[anXL3 crateNumber]);
+                [anXL3 setIsPollingXl3:false];
+            }
         }
-        if(xl3ispolling){
-            NSLog(@"Stopping XL3 polling \n");
-            [model stopXL3Polling];
-        }
-   }
+    }
 
     //Start the standard run and stop run initialization if failed
     if(![model startStandardRun:[model standardRunType] withVersion:[model standardRunVersion]]) return;
--- a/Source/Experiments/SNOP/SNOPController.m
+++ b/Source/Experiments/SNOP/SNOPController.m
@@ -87,6 +87,7 @@ snopGreenColor;
     doggy_icon = [[RunStatusIcon alloc] init];
     [self initializeUnits];
     return self;
+
 }
 - (void) dealloc
 {
@@ -223,7 +224,6 @@ snopGreenColor;
 
 }
 
-
 - (NSString*) defaultPrimaryMapFilePath
 {
     return @"~/SNOP";
@@ -258,9 +258,6 @@ snopGreenColor;
     [runControl getCurrentRunNumber]; //this should be done by the base class... but it is not
     //Sync SR with MTC
 
-    //Update XL3 state
-    [self updatexl3Mode:nil];
-
     [doggy_icon start_animation];
 
     [self initializeUnits];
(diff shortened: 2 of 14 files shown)


tlatorre-uchicago:
We should probably double check with Phil on the next call just to make sure he is ok with getting rid of the resync button since he originally came up with the restart/rollover/resync logic. I can't remember if he was present when we discussed it at the collaboration meeting.

jcaravaca:
This should be ready to go. I'll be testing it at the test-stand first and then at the detector next week. I think the new start/stop logic makes sense, but we should all agree that that's what we want.

jcaravaca:
Tested OK at the teststand and the detector. Once we figure what we want to do with the Resync logic we should be good to go.

Source/Experiments/SNOP/ECARun.m
```diff
@@ -142,8 +142,8 @@ - (void) setPulserRate:(NSNotification*)aNote
         NSArray *objs = [[(ORAppDelegate*)[NSApp delegate] document] collectObjectsOfClass:NSClassFromString(@"ORMTCModel")];
         if ([objs count]) {
             anMTCModel = [objs objectAtIndex:0];
-            [anMTCModel setPgtRate:[[self ECA_rate] floatValue]]; //UNCOMMENT THIS LINE BEFORE COMMISSIONING
-            [anMTCModel loadPulserRateToHardware];
+//            [anMTCModel setPgtRate:[[self ECA_rate] floatValue]]; //UNCOMMENT THIS LINE BEFORE COMMISSIONING
+//            [anMTCModel loadPulserRateToHardware]; //UNCOMMENT THIS LINE BEFORE COMMISSIONING
```
tlatorre-uchicago: Uncomment?

Source/Experiments/SNOP/SNOPController.m
```diff
@@ -486,60 +489,53 @@ - (void) SRVersionChanged:(NSNotification*)aNote
...
+    /* If we are not going to maintenance we shouldn't be polling */
+    unsigned long dbruntypeword = 0;
+    NSMutableDictionary* runSettings = [[[model standardRunCollection] objectForKey:[model standardRunType]] objectForKey:[model standardRunVersion]];
+    if(runSettings != nil){
+        //Get the run type word of the next run
+        dbruntypeword = [[runSettings valueForKey:@"run_type_word"] unsignedLongValue];
     }
-}
 
+    if( (dbruntypeword & kMaintenanceRun) || (dbruntypeword & kDiagnosticRun) ){
```
tlatorre-uchicago:
```
if (!((dbruntypeword & kMaintenanceRun) || (dbruntypeword & kDiagnosticRun)) {
    ...
```
will allow you to get rid of the superfluous block

Source/Experiments/SNOP/SNOPController.m
```diff
@@ -486,60 +489,53 @@ - (void) SRVersionChanged:(NSNotification*)aNote
...
+        //We are good
+    } else {
+        //Make sure we are not polling
+        NSArray* xl3s = [[(ORAppDelegate*)[NSApp delegate] document] collectObjectsOfClass:NSClassFromString(@"ORXL3Model")];
+        bool xl3ispolling = false;
+        for(ORXL3Model *anXL3 in xl3s){
+            if([anXL3 isPollingXl3]) xl3ispolling = true;
+        }
+        if(xl3ispolling){
+            BOOL cancel = ORRunAlertPanel(@"Not allowed to start a non-maintenance run with XL3s polling",@"Do you want to disable polling",@"No. Abort start run.",@"Yes. Disable and start run.",nil);
```
tlatorre-uchicago: Do we want to ask here? It seems like it would be fine to just disable polling.

Source/Experiments/SNOP/SNOPController.m
```diff
@@ -620,6 +617,9 @@ - (void) hvStatusChanged:(NSNotification*)aNote
             bool found;
             
             found = [hvStatusMatrix getRow:&mRow column:&mColumn ofCell:[hvStatusMatrix cellWithTag:[xl3 crateNumber]]];
+
+            NSLogColor([NSColor redColor], @"HV: row %d, crate %d, found %d", mRow, [xl3 crateNumber], found);
```
tlatorre-uchicago: I don't think we want to comment here every time the status changes.

Source/Experiments/SNOP/SNOPController.m
```diff
@@ -741,8 +768,60 @@ - (void) hvStatusChanged:(NSNotification*)aNote
...
+}
+
+- (void) triggerStatusChanged:(NSNotification*)aNote
+{
+    NSArray* xl3s = [[(ORAppDelegate*)[NSApp delegate] document] collectObjectsOfClass:NSClassFromString(@"ORXL3Model")];
+    for (ORXL3Model* xl3 in xl3s) {
+        //Take care of crate 17 and 18
+        int row = [xl3 crateNumber];
+        if (row > 16) row++;
+        [[triggerStatusMatrix cellAtRow:row column:0] setStringValue:[xl3 triggerStatus]];
```
tlatorre-uchicago: This will work currently, but I am updating the trigger stuff in my branch right now to get rid of the `triggerStatus` variable in the XL3 model. The problem is that it is redundant with the `isTriggerON` boolean. So, if you could change this to check `[xl3 isTriggerON]` instead and then just print `@"OFF"` or `@"ON"` that will make it easier for me when merging my changes.

Source/Experiments/SNOP/SNOPController.m
```diff
@@ -741,8 +768,60 @@ - (void) hvStatusChanged:(NSNotification*)aNote
...
+                [[globalxl3Mode cellAtRow:17 column:0] setStringValue:xl3ModeDescription];
+            }
+        }
+        else if ([anXl3 crateNumber] > 16){
+            [[globalxl3Mode cellAtRow:([anXl3 crateNumber]+1) column:0] setStringValue:xl3ModeDescription];
+        }
+        else{
+            [[globalxl3Mode cellAtRow:[anXl3 crateNumber] column:0] setStringValue:xl3ModeDescription];
+        }
+        //setStringValue:[xl3 xl3Mode] stringValue]];
```
tlatorre-uchicago: Delete the commented out lines?

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -800,12 +800,16 @@ - (void) runStarted:(NSNotification*)aNote
     }
 
     state = RUNNING;
-    [self setLastStandardRunType:[self standardRunType]];
-    [self setLastStandardRunVersion:[self standardRunVersion]];
-    [self setLastRunTypeWord:[self runTypeWord]];
-    NSString* _lastRunTypeWord = [NSString stringWithFormat:@"0x%X",(int)[self runTypeWord]];
-    [self setLastRunTypeWordHex:_lastRunTypeWord]; //FIXME: revisit if we go over 32 bits
+    if(start == COLD_START){
```
tlatorre-uchicago: Can you explain this change?

Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
```diff
@@ -2788,6 +2781,7 @@ - (void) writeXl3Mode: (uint32_t) mode withSlotMask: (uint32_t) slotMask
                              expectResponse:YES];
         NSLog(@"xl3 %02d set to %s mode.\n", [self crateNumber],
                 (xl3Mode == INIT_MODE) ? "INIT" : "NORMAL");
+        [[NSNotificationCenter defaultCenter] postNotificationName:ORXL3ModelStateChanged object:self];
```
tlatorre-uchicago: This is a low level function to write the mode which doesn't set the mode variable in the model, so this shouldn't post a notification. The two notifications you want to intercept are `ORXL3ModelStateChanged` and `ORXL3ModelXl3ModeChanged`.

Source/Objects/Custom Hardware/SNO+/XL3/XL3_LinkController.m
```diff
@@ -403,7 +403,17 @@ - (void) tabView:(NSTabView*)aTabView didSelectTabViewItem:(NSTabViewItem*)item
...
+    NSArray*  objs = [[(ORAppDelegate*)[NSApp delegate] document] collectObjectsOfClass:NSClassFromString(@"SNOPModel")];
+    SNOPModel* aSNOPModel = nil;
+    if ([objs count]) {
+        aSNOPModel = [objs objectAtIndex:0];
+    } else {
+        NSLogColor([NSColor redColor], @"xl3LockChanged: Couldn't find SNOP model. Please add it to the experiment and restart the run.\n");
+    }
 
+    BOOL notRunningOrInMaintenance = true; //default if no snopmodel found
+    if (aSNOPModel) notRunningOrInMaintenance = [aSNOPModel isNotRunningOrIsInMaintenance];
```
tlatorre-uchicago: This check is now possible without getting the SNO+ model right, i.e.

```
    return (![gOrcaGlobals runInProgress] ||
            (([gOrcaGlobals runType] & kMaintenanceRun) ||
            ([gOrcaGlobals runType] & kDiagnosticRun)));
```

Perhaps this should just be made a global function. There is no reason it should be a SNO+ model method.

Source/Objects/Custom Hardware/SNO+/XL3/XL3_Link.xib
```diff
@@ -81,6 +81,7 @@
                 <outlet property="hvRelayMaskMatrix" destination="4342" id="4408"/>
                 <outlet property="hvRelayOpenButton" destination="4065" id="4522"/>
                 <outlet property="hvRelayStatusField" destination="4322" id="4326"/>
+                <outlet property="hvRunStateLabel" destination="XZP-53-4IY" id="gon-Pw-7eU"/>
```
tlatorre-uchicago: I am currently editing the XL3 xib in my pull request. We'll have to figure out how to merge this.

Source/Experiments/SNOP/SNOPController.m
```diff
@@ -314,7 +305,17 @@ - (void) registerNotificationObservers
                      selector : @selector(hvStatusChanged:)
                          name : ORXL3ModelHVNominalVoltageChanged
                         object: nil];
-    
+
+    [notifyCenter addObserver : self
+                     selector : @selector(XL3ModeChanged:)
+                         name : ORXL3ModelStateChanged
```
tlatorre-uchicago: Should add a observer for `ORXL3ModelXl3ModeChanged`. See comment below.

Source/Experiments/SNOP/SNOPModel.m
jcaravaca: We only load new settings for cold starts, so the displayed run status (type, version, etc.) should change only in this case. This was a bug that was causing the displayed run status type and version to change when we rollover if the SR dropdown menu had changed.

Source/Objects/Custom Hardware/SNO+/XL3/XL3_LinkController.m
jcaravaca: Yeah, it's just a convenient function. I wrote it in the SNOP model since it's easier to get through than a global thing.

Source/Objects/Custom Hardware/SNO+/XL3/XL3_Link.xib
jcaravaca: What changes are you implementing? I just added two labels, so it shouldn't be hard to manually merging the changes.

jcaravaca:
Merged with master and re-tested OK at the detector.

Source/Objects/Custom Hardware/SNO+/XL3/XL3_LinkController.m
tlatorre-uchicago: I think it would be easier as a global function, because then we wouldn't need to add the complicated:

```
NSArray*  objs = [[(ORAppDelegate*)[NSApp delegate] document] collectObjectsOfClass:NSClassFromString(@"SNOPModel")];
SNOPModel* aSNOPModel = nil;
if ([objs count]) {
    aSNOPModel = [objs objectAtIndex:0];
} else {
    NSLogColor([NSColor redColor], @"xl3LockChanged: Couldn't find SNOP model. Please add it to the experiment and restart the run.\n");
}
```

in all the code.

tlatorre-uchicago:
I think the conclusion on the call yesterday was to leave the start/resync logic as it currently is due to the fact that we will have to turn triggers off every time we restart a run in this case.

jcaravaca:
I think I don't follow. The change brought here is that we never force a rollover because we thought that there is no use case for that and so, the 'Resync' would be the default when clicking 'Start/Restart'. Any problem with the 'Resync' functionality will be present also in the case we hit 'Resync' in the current master ORCA.

Source/Objects/Custom Hardware/SNO+/XL3/XL3_LinkController.m
jcaravaca: The problem here is that the OR globals don't know what a DiagnosticRun is, so we would need to implement our own globals.

tlatorre-uchicago: kMaintenanceRun is just an enum. It doesn't come from the SNO+ model. I'm just suggesting to make this a function and not a class method, not to create a function in the run control.

tlatorre-uchicago:
Yes, that's right. But a resync is equivalent to a hard stop and start. That means that the period in between the two the detector will not be live since all triggers will be turned off. I'm not sure what you mean about rollover, if I understand correctly you didn't change anything related to the rollover runs.

The original idea behind this was that the distinction between restart and resync was confusing, and it seemed that there wasn't a lot of difference in "physics" dead time between the two. However, as Phil pointed out, the difference is that the dead time during a continuous run start is still live (you still get triggers) but you just won't analyze it (it's in between first and valid gtid), while the dead time during a resync run is *actual* dead time, when the detector is blind to events since all the triggers are disabled.

I think this point is important, and I hadn't considered it when we discussed things at the collaboration meeting, but since we want to maximize our supernova live time I think it's worth keeping the distinction and just trying to educate operators about the difference. In addition (this is just a side benefit), the way things are currently there is no chance of the operator getting stuck with the buffers flushing when moving to a maintenance run if they hit restart. That will only happen when doing a resync or a hard run start.

jcaravaca:
OK, I missed the difference between 'continuous' and 'rollover'. I've brought back the Resync button, address your other comments and also lock the trigger interlock for physics, as I think it was the decision during yesterday's call.

Source/Experiments/SNOP/SNOPController.m
```diff
@@ -487,60 +495,49 @@ - (void) SRVersionChanged:(NSNotification*)aNote
...
 - (IBAction)resyncRunAction:(id)sender
 {
+
     /* A resync run does a hard stop and start without the user having to hit
      * stop run and then start run. Doing this resets the GTID, which resyncs
      * crate 9 after it goes out of sync :). */
 
-    //Load the standard run and stop run initialization if failed
-    if(![model loadStandardRun:[model standardRunType] withVersion:[model standardRunVersion]]) return;
-
```
tlatorre-uchicago: This needs to be brought back right?

Source/Experiments/SNOP/SNOPController.m
```diff
@@ -487,60 +495,49 @@ - (void) SRVersionChanged:(NSNotification*)aNote
...
     [model setResync:YES];
 
-    if ([[model document] isDocumentEdited]) {
-        [[model document] afterSaveDo:@selector(restartRun) withTarget:runControl];
-        [[model document] saveDocument:nil];
-    } else {
-        [runControl restartRun];
-    }
-}
```
tlatorre-uchicago: Same here.

Source/Experiments/SNOP/SNOPModel.m
tlatorre-uchicago: New settings are also loaded for continuous run starts.

Source/Objects/Custom Hardware/SNO+/XL3/XL3_LinkController.m
```diff
@@ -477,7 +471,9 @@ - (void) xl3LockChanged:(NSNotification*)aNotification
 
     //HV
     [self hvStatusChanged:nil];
-    
+    [hvTriggersButton setEnabled:!lockedOrNotRunningMaintenance];
+    [loadNominalSettingsButton setEnabled:!lockedOrNotRunningMaintenance];
```
tlatorre-uchicago: I think this also needs to be updated on line 865: https://github.com/jcaravaca/orca/blob/99bc86c81e2c9ac9be5a4c1dfd51171ebe6d6556/Source/Objects/Custom%20Hardware/SNO%2B/XL3/XL3_LinkController.m#L865.

Source/Experiments/SNOP/SNOPController.m
jcaravaca: This is now implicit in the startStandardRun method that is called below by the startRunAction

Source/Experiments/SNOP/SNOPController.m
jcaravaca: Also in the startStandardRun method

Source/Experiments/SNOP/SNOPController.m
```diff
@@ -1195,9 +1232,12 @@ - (IBAction) stopSmellieRunAction:(id)sender
...
 
     ////////////
-    // Roll over into maintinance run
+    // Roll over into maintenance run
     if([[model lastStandardRunType] isEqualToString:@"SMELLIE"]){
-        [model startStandardRun:@"MAINTENANCE" withVersion:@"DEFAULT"];
+        [model setStandardRunType:@"MAINTENANCE"];
+        [model loadStandardRun:[model standardRunType] withVersion:[model standardRunVersion]];
+        [model loadSettingsInHW];
+        [self startRunAction:self];
```
tlatorre-uchicago: `startRunAction` will call `startStandardRun` which will already load the standard run right? We don't want to change the hardware settings here since the run hasn't rolled over yet? Or am I missing something?

Source/Experiments/SNOP/SNOPController.m
```diff
@@ -1395,11 +1435,14 @@ - (IBAction) stopTellieRunAction:(id)sender
...
-    // Handle end of run sqeuencing
+    // Handle end of run sequencing
     if([[model lastStandardRunType] isEqualToString:@"TELLIE"]){
         // If user was running a TELLIE standard sequence, roll over into maintinance run
         if([self tellieStandardSequenceFlag]){
-            [model startStandardRun:@"MAINTENANCE" withVersion:@"DEFAULT"];
+            [model setStandardRunType:@"MAINTENANCE"];
+            [model loadStandardRun:[model standardRunType] withVersion:[model standardRunVersion]];
+            [model loadSettingsInHW];
+            [self startRunAction:self];
```
tlatorre-uchicago: Same here.

Source/Experiments/SNOP/SNOPController.m
jcaravaca: Mmh, I corrected this already... it might have gotten messed up with the rebases. Good catch, should be fine now.

tlatorre-uchicago:
merged.

PR #368, opened 2017-03-20
don't send a queue_run_start command for rollover runs

tlatorre-uchicago:
This commit updates the runInitialization method in the SNO+ model so that it
doesn't send the mtc server a queue_run_start command. This will cause the mtc
server to set the first and valid gtid fields in the run header to be the same,
which is how analysis will know that it was a rollover run.

I haven't tested this out yet, and it will also require the mtc server to be updated before a new release is made.


tlatorre-uchicago:
Tested at the teststand on March 20, 2017.

tlatorre-uchicago:
merged.

PR #371, opened 2017-03-22
ELLIE: Remove run number from db files index field

EdLeming:
This is a quick 2 line fix to remove the run number from the index field for database documents. This was a hangover used by old couchdb views and is incompatible with the new ratdb.


tlatorre-uchicago:
merged.

PR #372, opened 2017-03-25
SNO+: Single panic down -> Single ramp down

jcaravaca:
@BenLand100 would you ming checking that this would actually rampdown the crates, for both A and B supplies in case of crate 16?
@tlatorre-uchicago I'll test this at the TS in a minute.


BenLand100:
That will work for both A and B, yes.

jcaravaca:
The crate at the teststand is crate 19 and there is no button associated to this, so I'm afraid I can test it unless I hack my way around to it. In any case, I think we should test this at the detector before merging, so I'd vote for test it quickly tomorrow, if there is no other pending tests.

jcaravaca:
4/3/17 Tested OK at the detector. It'd be nice to have this merged soon so that I can continue the GUI development.

tlatorre-uchicago:
merged.

PR #373, opened 2017-03-27
Make changes to fix bugs in the TUBii gui for the speaker and counter masks.

icoulter:
Make changes to fix bugs in the TUBii gui for the speaker and counter masks. Namely, check boxes were numbered wrongly, ellie triggers were missing and the top 16 bits didn't get added to the mask properly.


tlatorre-uchicago:
merged.

PR #379, opened 2017-04-04
load nominal trigger settings from the database

tlatorre-uchicago:
This pull request adds a button to the XL3 model which will load the nominal trigger and sequencer settings from the database and load them to hardware.


tlatorre-uchicago:
Tested OK at the teststand. I verified that the new triggers on/off button behaves as expected and that the nominal states are fetched correctly when pressing the load nominal settings button.

tlatorre-uchicago:
merged.

PR #380, opened 2017-04-05
Add arrow key hot-keys to XL3, SNOCrate, SNORack, FEC and DB view

marzece:
Haven't tested yet. Probably will have to test on the full detector.

### DIFF
--- a/Source/Objects/Custom Hardware/SNO+/Fec32/ORFec32Controller.h
+++ b/Source/Objects/Custom Hardware/SNO+/Fec32/ORFec32Controller.h
@@ -143,6 +143,7 @@
 - (void) update20nTriggerInfo:(NSNotification*)aNote;
 - (void) update100nTriggerInfo:(NSNotification*)aNote;
 - (void) updateCmosReadInfo:(NSNotification*)aNote;
+- (void) keyDown:(NSEvent *)event;
 
 #pragma mark •••Actions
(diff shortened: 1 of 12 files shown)


marzece:
I briefly tested this UG. I tried sending a command from each of the dialogues I added hot-keys to and made sure the command got sent to the correct board. Everything seemed to work as expected.

tlatorre-uchicago:
Sweet! What are the cancelOperation functions for? I think it would also be good to add a bit more error checking. There seems to be a good example here: https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/EventOverview/HandlingKeyEvents/HandlingKeyEvents.html in Listing 5-4. Specifically, making sure that it has the right number of characters, ignoring modifier keys, and passing the keys onto the super class if it's not the key you're expecting.

marzece:
I made each chunk of code look a bit more like the thing you pointed to.

The cacelOperation is so when someone hits the ESC key it removes focus from the input field. For a couple of views if you started entering a value there was no way to 'exit' the input dialog, so the hot keys didn't work.

tlatorre-uchicago:
Looks like the crate view is different from the others, i.e. pressing h on crate view increases the crate number instead of decreasing it.

marzece:
Oops, fixed.

tlatorre-uchicago:
merged.

PR #381, opened 2017-04-05
TUBii: First commit of a TUBii version of PGT.

icoulter:
First commit of a TUBii version of PGT. Appears to work as expected.
Needs an accompanying update to TUBii's firmware.

### DIFF
--- a/Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiModel.m
+++ b/Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiModel.m
@@ -189,6 +189,10 @@ NSString* ORTubiiLock				= @"ORTubiiLock";
     NSString* const command = [NSString stringWithFormat:@"SetPrescaleTrigger %u %u",factor,mask];
     [self sendOkCmd:command];
 }
+- (void) setTUBiiPGT_Rate:(float)rate {
+    NSString* const command = [NSString stringWithFormat:@"SetTUBiiPGT %f",rate];
+    [self sendOkCmd:command];
+}
 
 - (void) setSmellieRate:(float)_rate {
     // Specifies the frequency (in Hz) that the smellie pulser will pulse at
--- a/Source/Objects/Custom Hardware/SNO+/TUBii/TUBii.xib
+++ b/Source/Objects/Custom Hardware/SNO+/TUBii/TUBii.xib
@@ -43,6 +43,7 @@
                 <outlet property="SpeakerMaskField" destination="AQh-yw-DbT" id="c9W-9d-4ub"/>
                 <outlet property="SpeakerMaskSelect_1" destination="6lV-eq-FNs" id="IyP-4D-Vv3"/>
                 <outlet property="SpeakerMaskSelect_2" destination="G7o-87-J17" id="tAD-S5-dpk"/>
+                <outlet property="TUBiiPGTRate" destination="0AE-vC-LbN" id="faN-DK-D8D"/>
                 <outlet property="TellieDelay_TextField" destination="Y6u-qe-A8u" id="oc4-cV-7UA"/>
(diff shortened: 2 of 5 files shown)


tlatorre-uchicago:
Lemme know when the firmware update is done, and I'll merge.

icoulter:
The testing of this didn't work due to a couple of issues (some related to this PR: I got two of the trigger bits swapped by mistake; some environmental: the firmware was expecting Ed's keep-alive pulse which hasn't been merged yet). We're going to test it again on Thursday with these issues fixed when the detector is at low voltage.

ecaden:
I'm not sure yet that we're ramping down on Thursday.

--
Erica Caden, Ph.D.
SNOLAB Research Scientist
SNO+ Detector Manager
ecaden@snolab.ca
+1.705.692.7000 x2253

icoulter:
Ah, your email on Friday said that there would be time available.

tlatorre-uchicago:
OK. I'll hold off until then. Why does the PGT depend on the keep alive?

icoulter:
It doesn't. I was rather silly and added the commits to the firmware in a linear way. I only realised after testing otherwise I could have worked around it.

icoulter:
Got the chance to test it now as we're on LV due to the lightning storm. There's one minor bug fix in the firmware that I half expected: the internal clock is at the wrong rate and can't be changed in the compiler I ended up sticking with which means the rate was slower than expected by a factor of 2. This is easy to fix in the firmware though. The ORCA PR is fine.

Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiModel.m
```diff
@@ -189,6 +189,10 @@ - (void) setPrescaleTrigger_Mask:(uint32_t)mask ByFactor:(uint32_t)factor {
     NSString* const command = [NSString stringWithFormat:@"SetPrescaleTrigger %u %u",factor,mask];
     [self sendOkCmd:command];
 }
+- (void) setTUBiiPGT_Rate:(float)rate {
+    NSString* const command = [NSString stringWithFormat:@"SetTUBiiPGT %f 999999999",rate];
```
tlatorre-uchicago: Just curious, what is the 999999999 for?

tlatorre-uchicago:
I think it would be nice to have the ability to set the pulser rate independently of enabling it. How does that work currently? The way the MTCD currently works you can set a rate for the pulser, and then enable/disable the pulser.

Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiModel.m
icoulter: Ah, forgot about this. The pulser needs a finite number of pulses to send. I'd hardcoded it to a large number.

icoulter:
Currently you choose a rate and then load the settings. To disable it you'd need to set the rate to zero and load or else just disable it from the trigger mask.

icoulter:
This should hopefully address your comments. I've changed the Send To TUBii button into two separate Start and Stop buttons.
I've also removed the hard coded number of pulses from ORCA into the firmware itself.

tlatorre-uchicago:
Please squash the commit into a single one with a commit message like:

```
SNO+: [description]

[longer description]
```

and I'll merge.

icoulter:
What's the command?

Quoting tlatorre-uchicago <notifications@github.com>:

tlatorre-uchicago:
The way I usually do it (I'm sure there is an easier way):

```
$ git checkout master
$ git checkout -b new_branch
$ git merge --squash TUBiiPGT
$ git commit
$ git push -f origin new_branch:TUBiiPGT
```

Might want to make sure you copy the whole directory before doing this in case it screws up.

icoulter:
Thanks. Looks like the squash worked.

tlatorre-uchicago:
merged.

PR #386, opened 2017-04-12
SNO+: TUBii unit corrections.

icoulter:
SNO+: TUBii previously used ns for the delay lengths and s for the pulse widths which was confusing. This brings things into line and changes the gui to give ns as units for everything.

### DIFF
--- a/Source/Objects/Custom Hardware/SNO+/ELLIE/ELLIEModel.m
+++ b/Source/Objects/Custom Hardware/SNO+/ELLIE/ELLIEModel.m
@@ -884,7 +884,7 @@ err:
             ////////////
             // Set the tubii model aand ask it to fire
             @try{
-                [theTubiiModel fireTelliePulser_rate:rate pulseWidth:100e-9 NPulses:[noShots intValue]];
+                [theTubiiModel fireTelliePulser_rate:rate pulseWidth:100 NPulses:[noShots intValue]];
             } @catch(NSException* e){
                 errorString = [NSString stringWithFormat:@"[TELLIE]: Problem setting TUBii parameters: %@\n", [e reason]];
                 NSLogColor([NSColor redColor], errorString);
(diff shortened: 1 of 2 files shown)


icoulter:
This is purely a change to gui as the rest is dealt with in the firmware.
@EdLeming Can you point me at the ELLIE code that needs to changed along with this?

EdLeming:
Hey @icoulter. Thanks for the heads up see below:

https://github.com/snoplus/orca/blob/master/Source/Objects/Custom%20Hardware/SNO%2B/ELLIE/ELLIEModel.m#L887

https://github.com/snoplus/orca/blob/master/Source/Objects/Custom%20Hardware/SNO%2B/ELLIE/ELLIEModel.m#L1858

Thanks!

icoulter:
Thanks.

tlatorre-uchicago:
Has this already been changed in the firmware?

icoulter:
No, I was waiting for some LV to do it as I don't like changing firmware while at HV. It shouldn't be merged until the firmware is updated though, otherwise ELLIE won't work properly.
Its been tested locally.

ecaden:
Can this be tested tomorrow then?

--
Erica Caden, Ph.D.
SNOLAB Research Scientist
SNO+ Detector Manager
ecaden@snolab.ca
+1.705.692.7000 x2253

icoulter:
Perhaps in the afternoon. I need to wait in for a package in the morning.
It doesn't really need testing as it worked locally and is a very  
minor change. Just a firmware update.

Quoting ecaden <notifications@github.com>:

> Can this be tested tomorrow then?
>
> --
> Erica Caden, Ph.D.
> SNOLAB Research Scientist
> SNO+ Detector Manager
> ecaden@snolab.ca
> +1.705.692.7000 x2253
>

icoulter:
New firmware has been loaded to the TUBii onsite.

tlatorre-uchicago:
merged. Note: Now that this has changed in the firmware, we should *not* run ELLIE until the new Orca release is finished.

icoulter:
Thanks.
Just as a comment, I don't think we can run ELLIE until the new release as it will send 5e-8 as the pulse width (expected it in seconds) which will effectively be 0 as the firmware works in nanoseconds now. In which case, it just won't do anything. So I think this mismatch is safe but you're right that we're probably best just not trying.

PR #387, opened 2017-04-14
Add nhit monitor to Orca

tlatorre-uchicago:
This pull requests adds a nhit monitor class to Orca. It is still a work in progress, but the idea is that you will be able to run the nhit monitor from the SNO+ control panel. You will also be able to set whether it should run automatically and on what crates it should automatically run on.


tlatorre-uchicago:
merged.

PR #388, opened 2017-04-17
	Return result if OK even if database is empty

boardhead:
Updates for Tony.  Not yet tested but changes were very simple.


tlatorre-uchicago:
merged.

PR #389, opened 2017-04-18
SNO+: Correct rack numbering typo in GUI

jcaravaca:
Just noticed a stupid typo in the rack numbering. I've been running master with this commit at the detector.


jcaravaca:
Just found out that the Resync button wasn't linked to the correct action. This has been tested at the detector.

tlatorre-uchicago:
merged.

PR #390, opened 2017-04-19
ECA Supernova-live and Physics-live modes and ECA campaign

jcaravaca:
- [x] Add ECA mode to allow to run with physics triggers enabled for low occupancy patterns
- [x] Add ECA campaign which allows to get several hardcoded ECA runs with one single click of the operator
- [x] Set the proper trigger word bits for PDST/TSLP automatically
- [x] Add ECA status display


jcaravaca:
This has been tested OK at the teststand. I wanted to include a status panel for the ECA runs as well, but that's orthogonal to the main functionality, which can be reviewed now.

jcaravaca:
Tested at the TS and detector May 11th. Should be good to go.

Source/Experiments/SNOP/ECARun.m
```diff
@@ -13,25 +13,46 @@
...
+NSString* ORECAStatusChangedNotification = @"ORECAStatusChangedNotification";
 NSString* ORECARunChangedNotification = @"ORECARunChangedNotification";
 NSString* ORECARunStartedNotification = @"ORECARunStartedNotification";
 NSString* ORECARunFinishedNotification = @"ORECARunFinishedNotification";
 
-//Definitions
+/* This flag must be enabled for commissioning 
+ * and normal detector running. Disable it only 
+ * for testing. */
+#define __COMMISSIONING__ 1
```
tlatorre-uchicago: I'm not a big fan of this. It seems like it's too easy to accidentally commit this to the main repository (for example, the embedded pedestals pull request currently has this commented out).

Perhaps you can create a separate branch for testing locally?

Source/Experiments/SNOP/ECARun.m
```diff
@@ -60,12 +81,112 @@ - (void) registerNotificationObservers
...
+            [previousSRVersion_campaign release];
+            eca_campaign_running = FALSE;
+            [[NSNotificationCenter defaultCenter] postNotificationName:ORECAStatusChangedNotification object: self userInfo: nil];
+            return;
+        }
+
+        //Rollover to previous run, if any.
+        if(start_new_run_campaign){
+            dispatch_sync(dispatch_get_main_queue(), ^{
+                [aSNOPModel setResync:YES];
```
tlatorre-uchicago: Why resync here?

Source/Experiments/SNOP/ECARun.m
```diff
@@ -60,12 +81,112 @@ - (void) registerNotificationObservers
...
+                [aSNOPModel startStandardRun:previousSR_campaign withVersion:previousSRVersion_campaign];
+            });
+        }
+        
+        //Don't leak memory
+        [previousSR_campaign release];
+        [previousSRVersion_campaign release];
+        eca_campaign_running = FALSE;
+
+        [[NSNotificationCenter defaultCenter] postNotificationName:ORECAStatusChangedNotification object: self userInfo: nil];
```
tlatorre-uchicago: This needs to be posted on the main thread.

Source/Experiments/SNOP/ECARun.m
```diff
@@ -60,12 +81,112 @@ - (void) registerNotificationObservers
...
+
+        dispatch_sync(dispatch_get_main_queue(), ^{
+            [self start];
+        });
+
+        //Wait until finished
+        while ([ECAThread isExecuting] || ![self isFinished]) {
+            usleep(1e4);
+        }
+        if(![ECAThread isExecuting] && [ECAThread isCancelled] && [self isFinished]){
```
tlatorre-uchicago: You are guaranteed at this point that `[ECAThread isExecuting]` is False and `[self isFinished]` is True right? It might make it easier to read if you just remove those from the check.

Source/Experiments/SNOP/ECARun.m
```diff
@@ -116,35 +269,70 @@ - (void) stop
...
+            unsigned long runTypeWord = [aSNOPModel runTypeWord];
+            runTypeWord &= ~(kECAPedestalRun & kECATSlopeRun); //Unset ECA bits
+            if ([ECA_type isEqualTo:@"PDST"]) {
+                runTypeWord |= kECAPedestalRun;
+            }
+            else if ([ECA_type isEqualTo:@"TSLP"]) {
+                runTypeWord |= kECATSlopeRun;
+            }
+            [aSNOPModel setRunTypeWord:runTypeWord];
         }
```
tlatorre-uchicago: Should print a warning here if you don't set anything.

Source/Experiments/SNOP/ECARun.m
```diff
@@ -116,35 +269,70 @@ - (void) stop
...
+            aSNOPModel = [objs objectAtIndex:0];
+            unsigned long runTypeWord = [aSNOPModel runTypeWord];
+            runTypeWord &= ~(kECAPedestalRun & kECATSlopeRun); //Unset ECA bits
+            if ([ECA_type isEqualTo:@"PDST"]) {
+                runTypeWord |= kECAPedestalRun;
+            }
+            else if ([ECA_type isEqualTo:@"TSLP"]) {
+                runTypeWord |= kECATSlopeRun;
+            }
+            [aSNOPModel setRunTypeWord:runTypeWord];
```
tlatorre-uchicago: Is this not a standard run? Why do we have to change the run type word here. This seems like it could cause problems, and will be very confusing to have the run type being changed outside of the standard runs.

Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
```diff
@@ -1850,46 +1866,11 @@ - (void) loadTriggers
...
-
-    if (ntohl(results->errorMask)) {
-        NSException *e = [NSException exceptionWithName:@"loadTriggersError"
-                          reason:@"failed to load channel triggers"
-                          userInfo:nil];
-        [e raise];
-    }
-}
-
-- (void) disableTriggers
```
tlatorre-uchicago: I'd prefer to keep this function as is since it might be useful. Could you just create a new function to AND the current triggers?

Source/Experiments/SNOP/ECARun.m
jcaravaca: I think this is safer than not having it, since it's very likely I forget to uncomment some line that was commented out and that'd be harder to catch that just this single line. If I created a separated branch for testing, at some point I have to bring the changes to the main one, and we will take the same risk.

Source/Experiments/SNOP/ECARun.m
jcaravaca: Not really since the previous piece of code ([self start] method) is non-blocking. It'll be running a parallel ECA thread and we should wait for it to finish.

Source/Experiments/SNOP/ECARun.m
jcaravaca: I thought we needed a stop/start at this point since we are going to a different SR. But maybe it's not really needed?

Source/Experiments/SNOP/ECARun.m
```diff
@@ -81,25 +202,47 @@ - (void) start
...
 
         //Set ECA Standard Run and start run:
         /* This needs to be outside the new thread since it will
          stop/start a new run, which would cancel the thread and
          exit after the first step... */
-        [aSNOPModel startStandardRun:@"ECA" withVersion:ECA_type];
+        BOOL runstarted = FALSE;
+        switch (ECA_mode){
+            case ECAMODE_DEDICATED:
+                [aSNOPModel setResync:YES];
```
jcaravaca: Same reason than above.

Source/Experiments/SNOP/ECARun.m
jcaravaca: We need to change the run type word for Supernova and Physics runs if we want to flag that it was an ECA run. The alternative is having 2 different versions for each SR with the different bit enabled, but that's even more dangerous since I'm sure we won't be able to keep them all in sync. I think the way I set the run type word here before the run actually start should be safe enough.

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -810,7 +810,7 @@ - (void) runStarted:(NSNotification*)aNote
 {
     ORRunModel *run = [aNote object];
 
-    uint32_t run_type = [run runType];
+    uint32_t run_type = [self runTypeWord];
```
jcaravaca: Because the SNOP runTypeWord is the one I'm changing during the ECAs and the changes might take a little to propagate to the runModel, resulting in a nasty race condition. I think we should always use our run type word, which actually is not currently true in some places.

Source/Experiments/SNOP/ECARun.m
tlatorre-uchicago: What part of it is non-blocking?

Source/Experiments/SNOP/ECARun.m
tlatorre-uchicago: Nope, no need to resync between standard runs. The resync is only needed during issues with synchronization.

Source/Experiments/SNOP/ECARun.m
tlatorre-uchicago: What is the mutually exclusive type for these runs? Don't we already have separate standard runs for PDST and TSLP or is this done the same way for those?

Source/Experiments/SNOP/SNOPModel.m
tlatorre-uchicago: There is no time to propagate. When notifications are posted they are immediately received and all functions which are observing them are run. I agree we should use a single run type word, but I think that should be the Orca run type since that is the one which controls the locks and is already used in most other places.

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -2064,7 +2064,7 @@ - (unsigned long) runTypeWord
 - (void) setRunTypeWord:(unsigned long)aValue
 {
     runTypeWord = aValue;
-    [[NSNotificationCenter defaultCenter] postNotificationName:ORSNOPRunTypeWordChangedNotification object: self];
+    [[NSNotificationCenter defaultCenter] postNotificationName:ORSNOPRunTypeWordChangedNotification object: nil];
```
tlatorre-uchicago: Why is this changed?

Source/Experiments/SNOP/ECARun.m
jcaravaca: [self start] launches the ECA thread which runs in parallel and the code will continue executing from there. So, we have to sleep until the first ECA is done to launch the second one.

jcaravaca:
Testing at the TS... I'll let you know when I make sure the latests changes do not break anything.

jcaravaca:
OK, everything keeps working fine, including the latest change on the run type word

tlatorre-uchicago:
I'm having a hard time following the flow of the new code, specifically about how the threads and run changes operate. Do you think you could either describe the basic flow or maybe produce a flow chart?

Specifically here are a few questions:

- What is a campaign?
- What does [ECARun start] do?
- What does [ECARun startCampaign] do?
- How many threads are running at a given time and how are they synchronized?

It would also be good to add a few comments to the top of these functions to describe what they do.

jcaravaca:
I'll add some more comments to further explain the logic. Find below a summary:
- startCampaign: this starts the startCampaignThread, which schedules 2 ECA runs: one pedestal and one time slopes. It does:
-- Set the correct settings for a pedestal run
-- Start a single eca run calling [ECARun start], which will launch the ECA thread in parallel
-- Wait until the latter is done
-- Set new settings for the time slope run
-- Wait until it is done
-- Restart the standard run that was running before the campaign started or stop the run if no run was ongoing
- start: this will start a new standard run for the ECAs to come. It sets a flag to True, allowing the doECAs method to be launched, which is the one that actually shoot the pedestals and contains the logic.
- launchECAThread: launches the actual ECA script (doECAs), which shoots the pedestals. This is done in this way to ensure we run AFTER the run as started.
- doECAs: set MTC settings in preparation for the ECAs and calls doPedestals or doTSlopes
- doPedestals/doTSlopes: contain the ECA algorithms

So, there are two threads:
- ECA thread: runs the actual algorithm for an individual ECA run.
- ECA campaign thread: launches ECA threads one by one, so that there is one one ECA thread running at a time (along with the ECA campaign thread that needs to monitor whether they finish, in order to launch the following).

tlatorre-uchicago:
Yeah, I think it would be good if you could add some documentation to explain all of them at the top of each function.

Currently, trying to follow the logic it seems like the flow is this:

- startCampaignThread is called which launches a thread
- this thread calls [self start]
- the start method starts a new run and sets a variable start_eca_run to True
- when the run start notification is posted, launchECAThread is called
- this method checks the start_eca_run variable and if it is set to True, launches doECAs in a separate thread
- original thread waits for doECAs to finish and repeats

In addition there are a few other places where internal variables are used to keep track of the state like eca_campaign_running. I think this is very likely to result in some weird race conditions and it's very hard to follow even with the explanations.

For example, currently the stop method only stops the main ECA thread. If stop is called in between the two eca runs in the campaign thread the thread will continue to run and start a new eca run. This specific example could be fixed by updating the stop method, but given the complexity of the rest of the logic, my guess is that there are more edge cases.

Would it be possible to simplify this to:

- register a function runStarted to be called when the run is started
- when the run is started, check the standard run and if it's an ECA run launch a new thread
- new thread either runs single TSLOPE, PDST, or both if it's a campaign run

If it's necessary to have these as separate runs then after the PDST or TSLOPE run it could launch a new run and set one variable telling it to run TSLOPE on the next.

This has the advantage that it's a single extra thread running and there is no need to synchronize things between multiple threads and the state is kept out of variables.

Could you also keep the existing loadTriggers and disableTriggers methods and just add a completely new function for the loadTriggersWithCrateMask? I realize it's a relatively minor change but these two existing functions are used elsewhere in the code and have been tested quite a bit.

jcaravaca:
I'll add some more documentation and inline describing the flow.

In few words: there is one thread that controls the actual ECA algorithm (PDST or TSLP) and another (if the campaign is run) that schedules the launching of the first one. The case you mention about stopping between runs during the campaign, would result in the campaign method returning and not launching the next run, since the thread was cancelled.

This was written in this way to profit from the already wrote method to run a single ECA run. I agree that there is a lot of room for optimization, but that'd probably require a major rewrite of the code... We could thoroughly test this PR at the detector taking few ECA runs to see if some weird edge case arise, but I didn't see anything potentially dangerous during my testing.

About the loadTriggers/disableTriggers, do you mean duplicating the code for each of the functions? They'll be the same with just one line difference...

tlatorre-uchicago:
Here is an example of what I mean:

Add the following line to the end of [ECARun start]:

```objective-c
[aSNOPModel startStandardRun:@"PHYSICS" withVersion:@"DEFAULT"];
```

If you run that on the teststand you will see that the ECA continues running in the middle of a physics run, including disabling channel triggers. This is the sort of race condition that I was talking about earlier.

These sorts of race conditions are very difficult to spot and even harder to debug, so in order to prevent them we need to enforce the following conventions:

- no thread which modifies the detector state should run across multiple runs
- try to minimize keeping state variables which control things at run boundaries as class member variables (i.e. start_eca_run, start_new_campaign_run, etc.)

Ideally everything should depend only on the run type word as well as any settings in the GUI and make sure that it cannot survive across run boundaries.

Here is a simple alternative:

1. At the start of a run, check the run type word. If it is an ECA PDST run, launch the appropriate thread, making sure that when runAboutToStop is called, it posts a run state change wait notification and kills the thread. If it is an ECA TSLP run, do the same. This eliminates the need for a "start_eca_run" variable. For example:

```objective-c

- (void) runStarted: (NSNotification *) aNote
{
    unsigned long runType = [gOrcaGlobals runType];

    if ([ecaThread isExecuting]) {
        NSLogColor([NSColor redColor], @"eca thread is already running!\n");
        [ecaThread cancel];
        return;
    }

    if (runType & kECARun) {
        [ecaThread initWithTarget:self selector:@selector(doECAs:) object:nil];
        [ecaThread start];
    }
}
```

I haven't gone through the run start sequence in detail, but I guess one reason you implemented things this way is so that you have a chance to change the run type word before the run starts. This can be fixed by simply creating two standard runs: ECA-PDST and ECA-TSLP, or creating two run versions (which will be possible for operators to run with the next release).

2. isFinished can be replaced by ![ECAThread isRunning], i.e.

```objective-c
- (BOOL) isFinished
{
    return ![ecaThread isExecuting];
}
```

3. For the campaign, have a single class variable which indicates that after a pdst run to run a tslp run. Then, at the end of a successful PDST run, check this variable, and if it's set restart into a tslp run.

One alternative to 3 above which would generalize a bit nicer, would be to have a variable in the SNOPModel class which indicates which standard run should be run when the run starts. Then, instead of loading the standard run values to the GUI in `startStandardRun` we could do so in `runAboutToStart`. Then, all the ECA class would have to do at the end of a successful eca run if it is in campaign mode is to set this variable to an eca pdst run, i.e.

```objective-c
- (void) doECAs
{
    /* Do ECA stuff. */
    if (campaign) {
        [aSNOPModel setNextStandardRun:@"ECA" withVersion:@"TSLP"];
    } else {
        [aSNOPModel setNextStandardRun:@"PHYSICS" withVersion:@"DEFAULT"];
    }

    return;

err:
    /* If there was an error or we were stopped, then don't change the next standard run. */
    return;
}
```

If you're interested in the last option, which I think might be the best, I'm happy to update the SNO+ model to work this way and push it up, just let me know.

Regarding the loadTriggers, yes I would prefer to have it copied and not changed, even though it is just a few lines. Again, these functions are used extensively and have been tested thoroughly. I realize in general it's good to reuse code by calling functions, but I'd prefer not to have this changed since this hasn't been tested as thoroughly. For example, you call `[[aXL3Mask objectAtIndex:slot] unsigned IntValue]`, which will probably work on the Macs since I think ints are 32 bits there, but that isn't guaranteed and on some machines it could be 2 bytes, in which case it would only and the first 2 bytes of the trigger mask.

jcaravaca:
If ECAs are run in the middle of a physics run, they won't change the trigger masks, since we would be in 'Physics' mode, that doesn't allow that.

The problem with (1) is that we don't want to run the ECA algorithm when the ECA standard run is choses. It'd be confusing to start another thread and an entire background process just because the selected standard run has the ECA bit set. Also, we want to allow ECAs in Supernova and Physics, so in principle we should run the thread for runs with the PDST or TSLP bit set.
Previously, I had two different ECA standard runs, one for PDSTs and another for TSLPs, with different bit sets. The problem is that now we want to allow supernova runs to have those bit sets, and it's safer to have a default one and set the bits afterwards than having 2 standard runs that need to be kept in sync. That's why I moved to this implementation.

(2) I had loads of problems with this due to the thread being always allocated in memory. That led to [ecaThread isExecuting] always being 'False' when the second ECA is run. I think I found the solution to this and I might be able to implement your approach.

(3) Since I'm not sure (1) is the best option, I don't think your second approach would work. I agree on the first one though. I'll work to remove the campaign thread.

jcaravaca:
After a look to the code:
(2) it's gonna be trickier than what I thought -- the problem is that the ECAThread always exists, meaning that if I run an ECA run, stop and run the next one, the statement [ECAThread isFinished] will be true before we launch the new thread. I couldn't come up with a better solution that implementing the variable 'isFinished'. If you have a better idea, let me know.
(3) It's not so simple since the ECAs depend on different parameters that need to be set before starting the ECA algorithm. So the logic will be way more complicated since I have to launch a new ECA run within the first ECA thread, once the run started change the ECA parameters and finally start the ECA logic. That's pretty much what the campaign does in a clearer way, according to me. Again, suggestions welcome.

tlatorre-uchicago:
> If ECAs are run in the middle of a physics run, they won't change the trigger masks, since we would be in 'Physics' mode, that doesn't allow that.

In the example I tried, I was just running a standard ECA run, so I don't think it was running in physics mode.

> The problem with (1) is that we don't want to run the ECA algorithm when the ECA standard run is choses. It'd be confusing to start another thread and an entire background process just because the selected standard run has the ECA bit set. Also, we want to allow ECAs in Supernova and Physics, so in principle we should run the thread for runs with the PDST or TSLP bit set.

I'm not sure I understand what you mean when you say you don't want to run the ECA algorithm when the ECA standard run is chosen. The current ECA code already launches a thread when an ECA is run, so perhaps I'm misunderstanding you here.

> Previously, I had two different ECA standard runs, one for PDSTs and another for TSLPs, with different bit sets. The problem is that now we want to allow supernova runs to have those bit sets, and it's safer to have a default one and set the bits afterwards than having 2 standard runs that need to be kept in sync. That's why I moved to this implementation.

I see your point here. Essentially what we want is something extra to the standard run to specify these bits in the run word.

Currently there are a plethora of options which makes it a little bit confusing:

1. Select ECA run from the standard runs and start which currently won't do anything (I think?)
2. Press Start Single ECA Run, which will run a single ECA run with the settings in the GUI as the parameters
3. Press Start ECA Campaign which will run as supernova live regardless of the GUI and use hardcoded values instead of the GUI values.

One solution to this is to have the ECA thread start regardless of any of these buttons if any of the ECA bits are masked in in the run word, then it could take appropriate action based on both the run word and the GUI values. For example, there could be a radio box to check whether you want to use the standard parameters or the paramters in the GUI (or a single button to reset the parameters to the nominal values). Then, the dedicated, supernova, and physics live versions would all just depend on the run type word, i.e. if the run type is ECA then do a dedicated run, if the run type is physics but the ECA PDST bit is checked then do a physics live run, etc. There could be another checkbox that says "Run TSLP after pedestal" which combined with the "use default values" would essentially become the campaign mode if you ran a supernova run with the ECAPDST bit checked.

Creating separate standard runs is an easy way to achieve this, but it does come with the burden of making sure the settings are the same. Is it possible to save standard runs outside of a maintenance run? In that case, it would be very easy to make sure they are in sync. Alternatively, a web interface showing the standard run values would also make it easy to check.

> (2) it's gonna be trickier than what I thought -- the problem is that the ECAThread always exists, meaning that if I run an ECA run, stop and run the next one, the statement [ECAThread isFinished] will be true before we launch the new thread. I couldn't come up with a better solution that implementing the variable 'isFinished'. If you have a better idea, let me know.

Shouldn't there only be one thread per run now?

> (3) It's not so simple since the ECAs depend on different parameters that need to be set before starting the ECA algorithm. So the logic will be way more complicated since I have to launch a new ECA run within the first ECA thread, once the run started change the ECA parameters and finally start the ECA logic. That's pretty much what the campaign does in a clearer way, according to me. Again, suggestions welcome.

Is the problem that at the start of the run, you don't know whether it's a campaign or not? Otherwise, I don't think I understand the problem. In the case of the campaign all of the settings are hardcoded, so you just load those and then do either the pedestals or time slopes, and if not you get them from the model variables which come from the GUI. Perhaps I'm misunderstanding something though.

tlatorre-uchicago:
Thinking a little more about this, shouldn't it be possible to do everything based on the run type word? I mean something like this:

```objective-c
- (void) runAboutToStart: (NSNotification *) aNote
{
    uint32_t runType = [gOrcaGlobals runType];

    if (runType & (kECAPedestalRun | kECATSlopeRun)) {
        [ecaThread initWithTarget:self selector:@selector(doECAs:) object:nil];
    }
}

- (void) doECAs
{
    uint32_t runType = [gOrcaGlobals runType];

    if (runType & kPhysicsRun) {
        /* This is a physics ECA run, so use the standard ECA settings. */
    } else if (runType & kSupernovaRun) {
        /* This is a supernova ECA run, so use the standard ECA settings. */
    } else if (runType & kECARun) {
        /* This is a dedicated ECA run, so use the GUI settings for the parameters. */
    }

    if (runType & kECAPedestalRun) {
        /* Do pedestals. */
    } else if (runType & kECATSlopeRun) {
        /* Do time slopes. */
    }

    if ((runType & (kSupernovaRun | kECAPedestalRun)) == (kSupernovaRun | kECAPedestalRun)) {
        /* Just finished a successful supernova PDST run, so we restart into a TSLP run. */
        [aSNOPModel setNextStandardRun:@"SUPERNOVA-ECA-TSLP" withVersion:@"DEFAULT"];
    }

    return;

err:

    return;
```

I agree that this has the disadvantage of needing to create several different standard runs with all the same settings, but I think it will make the system a lot more clear and robust. We could also think about doing something like making the standard run name refer only to hardware settings, and the standard run version change the run type word. That would solve this problem, but I guess would require a change to the backend.

jcaravaca:
It's true that using the runTypeWord to determine whether taking ECAs would be a lot cleaner, but I'm a bit hesitant on having ORCA launching threads with some more or less complicated logic based on a single check box on the GUI. I prefer to have a button that explicitly launches the ECA algorithm, otherwise seems to obscure to me. The Standard Runs were designed to put the detector in a known state, so maybe this is overkill. Also, I'll have to entirely change the logic, which we know has been working in the past months (the only new thing would be the ECA campaign). Can we set a time to discuss this over the phone and try to reach a compromise? We can name the main potential problems and I'll try to address them with minimal changes.

tlatorre-uchicago:
> It's true that using the runTypeWord to determine whether taking ECAs would be a lot cleaner, but I'm a bit hesitant about having ORCA launching threads with some more or less complicated logic based on a single check box on the GUI.

But the check box is the sole thing which determines the run type, i.e. if the ECA bit is checked in the run type word, then everything downstream is going to treat the run like an ECA run, and if it's not checked then it will be treated like a normal run. Launching the ECA algorithm based on the run type word guarantees that all runs marked as ECA runs are ECA runs and vice versa.

> I prefer to have a button that explicitly launches the ECA algorithm, otherwise seems too obscure to me. The Standard Runs were designed to put the detector in a known state, so maybe this is overkill.

I imagined that the standard runs were designed exactly for this purpose, i.e. to allow the operators to do runs by simply selecting a run and pressing start. Having different buttons which start/stop runs outside of the main three in the SNO+ control panel seems like it would be more confusing.

> Can we set a time to discuss this over the phone and try to reach a compromise? We can name the main potential problems and I'll try to address them with minimal changes.

Sure, I'm available most of this week, just send me an email and we can set something up.

tlatorre-uchicago:
Superseded by #454.

PR #391, opened 2017-04-20
SNO+: update SNO+ model and run model to load elapsed time since run start

tlatorre-uchicago:
This commit updates the SNO+ model so that it loads the elapsed time since the
run started from the detector state database to the run model on boot up. This
required a small change to the run control to start the run timer so that the
elapsed time is incremented properly.

With this update, the run handover process should not require a new run to be started. I am not going to merge this until getting approval from Mark Howe since I had to edit the run control object but I will keep it in the new release.


tlatorre-uchicago:
merged.

PR #398, opened 2017-05-02
Check relays

marzece:
This adds two things. Firstly the check interlock button is now check interlock & relays*. Second, when turning HV on the the following things are checked. Interlock, XL3 mode (init/normal), HV relays are known, and HV relays are open for any missing fecs. If any of those fail the HV won't turn on.

Additionally this will require ML403 code that's as recent or better than [this](https://github.com/pennneutrinos/fullml403/commit/9c7d20967b355fa97ff07267ac96887516e12b09) commit.
So we should probably consider versioning and documenting that dependency somewhere and stuff like that.

@BenLand100 would you mind taking a look at this and making sure it won't disrupt any of the existing HV code.

*Not actually, my macbook is several version of xcode too advanced so I haven't made any .xib changes. I'll have to change the xl3 gui using one of the site computer when test it or something.

### DIFF
--- a/Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
+++ b/Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
@@ -4322,6 +4322,8 @@ err:
                 // Finally make sure the XL3s are reading out
                 if([xl3 xl3Mode] == NORMAL_MODE) {
                     goodModes++;
+                } else {
+                    NSLogColor([NSColor redColor],@"%@ NOT in normal mode\n",[[xl3 xl3Link] crateName]);
                 }
             }
         }
--- a/Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
+++ b/Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
@@ -4238,33 +4238,36 @@ err:
 
     NSLog(@"%@ switch A is %@, switch B is %@.\n",[[self xl3Link] crateName], switchAIsOn?@"ON":@"OFF", switchBIsOn?@"ON":@"OFF");
 }
+- (uint32_t) checkRelays:(uint64_t)relays {
+    // Returns if the relays are acceptable or not.
+    // Currently just checks that any missing slots also have open relays
+    // Could be expanded in the future.
 
-- (void) setHVSwitch:(BOOL)aOn forPowerSupply:(unsigned char)sup
-{
-    @synchronized(self) {
-    BOOL xl3SwitchA, xl3SwitchB;
-
-    @try {
-        [self readHVSwitchOnForA:&xl3SwitchA forB:&xl3SwitchB];
+    uint32_t bad_slots = 0;
+    uint32_t slots = [self getSlotsPresent];
+    for (int i=0;i<16;i++) {
+        if((slots & 1<<16) == 0) {
+            if(((0xF << 1*4) & relays) != 0)
+                bad_slots += 1<<i;
+        }
     }
-    @catch (NSException *exception) {
-        NSLog(@"%@ error in readHVSwitch\n", [[self xl3Link] crateName]);
-        return;
-    }
-
-    [self setHvASwitch:xl3SwitchA];
-    [self setHvBSwitch:xl3SwitchB];
-        
-        
+    return bad_slots;
+}
+- (BOOL) isHVAdvisable:(unsigned char) sup {
     BOOL interlockIsGood = false;
     BOOL relaysKnown = false;
+    BOOL relaysGood = true; // Start true b/c if relaysKnown doesn't pass this won't be checked.
+    BOOL modeGood = false;
     uint64_t relays;
     if ([self crateNumber] == 16 && sup != 0) { //16B
         //Checks interlocks for any crates with connected OWL tubes
         //Assumes that all relevant crates are represented in the open experiment
         uint32_t checkedCrates = 0;
         uint32_t goodInterlocks = 0;
+        uint32_t knownRelays = 0;
         uint32_t goodRelays = 0;
+        uint32_t goodModes = 0;
+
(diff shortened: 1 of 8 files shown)


marzece:
I tested this on the penn teststand and it worked well. Obviously I wasn't able to test the OWL related code. I think that will have to wait till I'm able to try it on the detector

BenLand100:
:+1: nothing here should cause problems elsewhere.

Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
```diff
@@ -4114,6 +4114,39 @@ - (void) setHVRelays:(unsigned long long)aRelayMask
...
+    uint32_t mask1, mask2, known;
+
+    memset(payload, 0, XL3_PAYLOAD_SIZE);
+    GetHVRelaysResults* data = (GetHVRelaysResults*) payload;
+
+    @try {
+        [[self xl3Link] sendCommand:GET_HV_RELAYS_ID withPayload:payload expectResponse:YES];
+    }
+    @catch (NSException *exception) {
+        NSLogColor([NSColor redColor],@"%@ error sending GetHVRelays command.\n",[[self xl3Link] crateName]);
```
tlatorre-uchicago: Should print the exception name and reason here. Something like:

```
NSLogColor([NSColor redColor], @"crate %02d: error getting HV relay state. error: %@ reason: %@", [self crateNumber], [e name], [e reason]);
```

If you are going to let the exception bubble up though, it's probably just best to not wrap this in a try except block.

Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
```diff
@@ -4213,37 +4246,50 @@ - (void) readHVSwitchOn
...
-    BOOL xl3SwitchA, xl3SwitchB;
-
-    @try {
-        [self readHVSwitchOnForA:&xl3SwitchA forB:&xl3SwitchB];
-    }
-    @catch (NSException *exception) {
-        NSLog(@"%@ error in readHVSwitch\n", [[self xl3Link] crateName]);
-        return;
+- (uint32_t) checkRelays:(uint64_t)relays {
+    // Returns if the relays are acceptable or not.
```
tlatorre-uchicago: I would add to this to say exactly what it returns. In this case it looks like it is returning a bitmask of any slots which have relays closed for cards which aren't present.

Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
```diff
@@ -4213,37 +4246,50 @@ - (void) readHVSwitchOn
...
+    uint32_t bad_slots = 0;
+    uint32_t slots = [self getSlotsPresent];
+    for (int i=0;i<16;i++) {
+        // If slot is missing
+        if((slots & 1<<i) == 0) {
+            // And if the 4 relays for that slot aren't all 0 (open)
+            uint64_t mask = (uint64_t)0xF << i*4;
+            if((mask & relays) != 0)
+                //Then set a bit in bad_slots
+                bad_slots += 1<<i;
```
tlatorre-uchicago: Maybe change this to  ```bad_slots |= 1 << i;``` in case another check is added later.

Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
```diff
@@ -4254,20 +4300,97 @@ - (void) setHVSwitch:(BOOL)aOn forPowerSupply:(unsigned char)sup
...
+                    [xl3 readHVRelays:&relays isKnown:&good];
+                    if (good) {
+                        knownRelays++;
+                        // If the relays are known make sure slot 15's are open if the FEC is missing
+                        uint32_t bad_relays = [xl3 checkRelays:relays];
+                        if((bad_relays & 1<<15) == 0) {
+                            goodRelays++;
+                        }
+                        else {
+                            NSLogColor([NSColor redColor], @"%@ HV relays for slot 15 are closed but FEC is missing!\n");
```
tlatorre-uchicago: missing a argument to NSLogColor

Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
```diff
@@ -4254,20 +4300,97 @@ - (void) setHVSwitch:(BOOL)aOn forPowerSupply:(unsigned char)sup
...
+                        }
+
+                    } else {
+                        NSLogColor([NSColor redColor],@"%@ HV Relays unknown\n",[[xl3 xl3Link] crateName]);
+                    }
+                }
+                @catch (NSException *exception) {
+                    NSLogColor([NSColor redColor],@"%@ error in readHVRelays\n",[[xl3 xl3Link] crateName]);
```
tlatorre-uchicago: print exception name and reason

Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
```diff
@@ -4254,20 +4300,97 @@ - (void) setHVSwitch:(BOOL)aOn forPowerSupply:(unsigned char)sup
...
         }
+
+        @try {
+            [self readHVRelays:&relays isKnown:&relaysKnown];
+            if (!relaysKnown){
+                NSLogColor([NSColor redColor],@"%@ HV relays unknown\n",[[self xl3Link] crateName]);
+            }
+        }
+        @catch (NSException *exception) {
+            NSLogColor([NSColor redColor],@"%@ error in readHVRelays\n",[[self xl3Link] crateName]);
```
tlatorre-uchicago: print exception name and reason

marzece:
I tested this on the detector earlier today. It seemed to work as expected with the exception of a missing log message which I've since added. This should be good to go

tlatorre-uchicago:
merged.

PR #399, opened 2017-05-04
Embedded pedestals

jcaravaca:
Add functionality to inject pedestals at a fixed rate of 1Hz and using the channel-by-channel ECA pattern in any run but ECA runs. Take into account if the NHitMonitor is running as well.


jcaravaca:
Tested at the teststand 5/3/17. It should be tested at the detector.

jcaravaca:
Tested at the TS and the detector May 11th.

Source/Experiments/SNOP/ECAPatterns.h
```diff
@@ -17,3261 +17,17 @@
 #ifndef Orca_ECAPatterns_h
 #define Orca_ECAPatterns_h
 
+#define NCRATES 19
+#define NCARDS 16
```
tlatorre-uchicago: I think it's probably worth using a more specific name so these names could potentially be used elsewhere. Alternatively, you could make them global static variables.

Source/Experiments/SNOP/LivePedestals.m
```diff
@@ -0,0 +1,384 @@
...
+#import "ORFec32Model.h"
+
+//Notifications
+NSString* ORLivePedsChangedNotification = @"ORLivePedsChangedNotification";
+
+
+/* This flag must be enabled for commissioning
+ * and normal detector running. Disable it only
+ * for testing. */
+//#define __COMMISSIONING__ 1
```
tlatorre-uchicago: This shouldn't be commented right?

Source/Experiments/SNOP/LivePedestals.m
```diff
@@ -0,0 +1,384 @@
...
+
+#define PEDGT_DELAY 150 //ns
```
tlatorre-uchicago: I'd give this a more specific name or make it a global static variable.

Source/Experiments/SNOP/LivePedestals.m
```diff
@@ -0,0 +1,384 @@
...
+    run_livepeds = FALSE;
+}
+
+- (void) launchLivePedsThread:(NSNotification*)aNote
+{
+
+    /* We always launch the LivePed thread. It will check periodically
+     * whether we should send pedestals or not, based on the radio
+     * button and the NHit monitor. */
+    if(![PedThread isExecuting]){
```
tlatorre-uchicago: I don't think we want the pedestals running during all run types. This should only happen during physics running. Also, I think it's probably better to centralize this and have it called from the SNO+ model as opposed to catching the run start notifications.

Source/Experiments/SNOP/LivePedestals.m
```diff
@@ -0,0 +1,384 @@
...
+                prev_finedelay = [anMTCModel fineDelay];
+                prev_pedmask = [anMTCModel pedCrateMask];
+                prev_pulserRate = [anMTCModel pgtRate];
+
+                // Set correct MTC settings for the pedestals
+                [anMTCModel setCoarseDelay:PEDGT_DELAY];
+                [anMTCModel setFineDelay:0];
+                [anMTCModel setGtMask:([anMTCModel gtMask] | MTC_EXT_8_MASK)];
+                [anMTCModel setIsPedestalEnabledInCSR:YES];
+                [anMTCModel setPgtRate:1]; //1Hz
```
tlatorre-uchicago: Need to be run on the main thread. Might be a good idea to have the PGT rate as a user defined number in case we want to change it in the future.

Source/Experiments/SNOP/LivePedestals.m
```diff
@@ -0,0 +1,384 @@
...
+
+                // Set EPED records
+                [aSNOPModel updateEPEDStructWithCoarseDelay:[anMTCModel coarseDelay] fineDelay:[anMTCModel fineDelay] chargePulseAmp:0x0 pedestalWidth:[anMTCModel pedestalWidth] calType:30];
```
tlatorre-uchicago: Also needs to happen on the main thread.

Source/Experiments/SNOP/LivePedestals.m
```diff
@@ -0,0 +1,384 @@
...
+                    [aSNOPModel updateEPEDStructWithStepNumber:current_step];
+                    dispatch_sync(dispatch_get_main_queue(), ^{
+                        [aSNOPModel shipEPEDRecord];
+                    });
+
+                    //Fire one pedestal, hopefully...
+                    dispatch_sync(dispatch_get_main_queue(), ^{
+                        [anMTCModel continueMTCPedestalsFixedRate];
+                    });
+                    usleep(1e6); //sleep for a second while firing pedestals
```
tlatorre-uchicago: If the pedestal rate is 1 Hz shouldn't you sleep longer than 1 second to make sure you get a trigger?

Source/Experiments/SNOP/LivePedestals.m
```diff
@@ -0,0 +1,384 @@
...
+                    }
+
+                } //end while loop
+
+                // Recover previous MTC settings
+                [anMTCModel setCoarseDelay:prev_coarsedelay];
+                [anMTCModel setFineDelay:prev_finedelay];
+                [anMTCModel setGtMask:~(~[anMTCModel gtMask] | MTC_EXT_8_MASK)];
+                [anMTCModel setIsPedestalEnabledInCSR:NO];
+                [anMTCModel setPgtRate:prev_pulserRate];
```
tlatorre-uchicago: Needs to happen on the main thread.

Source/Experiments/SNOP/LivePedestals.m
```diff
@@ -0,0 +1,384 @@
...
+    stop:
+
+        current_status=LIVEPEDS_STOPPED;
+
+        // Recover previous MTC settings
+        [anMTCModel setCoarseDelay:prev_coarsedelay];
+        [anMTCModel setFineDelay:prev_finedelay];
+        [anMTCModel setGtMask:~(~[anMTCModel gtMask] | MTC_EXT_8_MASK)];
+        [anMTCModel setIsPedestalEnabledInCSR:NO];
+        [anMTCModel setPgtRate:prev_pulserRate];
```
tlatorre-uchicago: Needs to happen on the main thread.

Source/Experiments/SNOP/LivePedestals.m
```diff
@@ -0,0 +1,384 @@
...
+        dispatch_async(dispatch_get_main_queue(), ^{
+            [fec setPedEnabledMask:mask];
+        });
+    }
+
+    //Load pedestal mask
+    for (ORXL3Model *xl3 in anXL3Model) {
+        //NSLog(@"Set pedestal for XL3 %d \n", [xl3 crateNumber]);
+        dispatch_async(dispatch_get_main_queue(), ^{
+            [xl3 setPedestalInParallel];
```
tlatorre-uchicago: This doesn't have to be done on the main thread since it just launches a new thread.

Source/Experiments/SNOP/LivePedestals.m
```diff
@@ -0,0 +1,384 @@
...
+
+- (void)setCurrent_step:(int)aValue
+{
+    current_step = aValue;
+    [[NSNotificationCenter defaultCenter] postNotificationName:ORLivePedsChangedNotification object:@{@"step":[NSNumber numberWithInt:aValue]}];
+}
+
+- (bool) isRunning
+{
+    return (current_status==LIVEPEDS_RUNNING);
```
tlatorre-uchicago: Can this just check whether the thread is running?

Source/Experiments/SNOP/NHitMonitor.m
```diff
@@ -491,6 +491,11 @@ - (void) run: (NSDictionary *) args
             goto err;
         }
 
+        //Set EPED record
+        //Mark NHit monitor initialization with an EPED record
+        [snop updateEPEDStructWithCoarseDelay:COARSE_DELAY fineDelay:FINE_DELAY chargePulseAmp:0x0 pedestalWidth:PEDESTAL_WIDTH calType:40];
+        [snop shipEPEDRecord];
```
tlatorre-uchicago: Needs to happen on the main thread.

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -478,25 +482,25 @@ - (void) registerNotificationObservers
...
+//
+//    [notifyCenter addObserver : self
+//                     selector : @selector(runAboutToStart:)
+//                         name : ORRunSecondChanceForWait
+//                       object : nil];
+//
+//    [notifyCenter addObserver : self
+//                     selector : @selector(runStarted:)
+//                         name : ORRunStartedNotification
+//                       object : nil];
```
tlatorre-uchicago: Why are these commented out?

Source/Objects/Custom Hardware/SNO+/MTC/ORMTCModel.m
```diff
@@ -1625,7 +1625,6 @@ - (void) loadCoarseDelayToHardware
 
     @try {
         [mtc okCommand:"set_coarse_delay %i", coarse_delay];
-        NSLog(@"mtc: coarse delay %ins\n", coarse_delay);
```
tlatorre-uchicago: These are very useful when setting settings on the MTC. I think at a minimum we should add printouts to the controller so you know what you are setting, although this still won't work for when the whole MTC is initialized. How often are the settings changed during embedded pedestals?

Source/Experiments/SNOP/LivePedestals.m
jcaravaca: Good catch.

Source/Experiments/SNOP/LivePedestals.m
jcaravaca: We need to monitor the pedestals also during calibration runs to correct by the different trigger rate (if a correction is actually needed at the end...), e.g. embedded pedestals will be very useful for PCA and N16 runs.
I like the way it's implemented now since it's more modular. I agree that it's not centralized and it could be hard to follow the entire logic of the run start/stop, but I don't see any potential failure mode the way it is. Do you have any concern other than not having everything centralized?

Source/Experiments/SNOP/LivePedestals.m
jcaravaca: I'll ship this to the main thread. I actually prefer having the parameters hardcoded since we don't want to run at a very high pulser rate. Otherwise, the actual pedestaling will interfere with the actual measurement. I guess I'm worried about someone changing it and running at higher rate over few hours, invalidating the actual measurement.

jcaravaca: Oh, now that I look into this: does this need to happen in the main thread? I'm just setting values in the model, not sending anything to HW.

Source/Experiments/SNOP/LivePedestals.m
jcaravaca: Same than above: is that actually needed?

Source/Experiments/SNOP/LivePedestals.m
jcaravaca: Yup, good point.

Source/Experiments/SNOP/LivePedestals.m
jcaravaca: The LivePed thread is actually always running when the run is ongoing, and it continuously checks whether conditions meet to start shooting pedestals.

jcaravaca:
@tlatorre-uchicago I've noticed that I didn't push properly the last commits from the teststand and that's why there are some commented-out code and such. I'll push the latest commits plus amendments for your comments.

Source/Objects/Custom Hardware/SNO+/MTC/ORMTCModel.m
jcaravaca: This is a pain for both ECAs and LivePeds, since we usually change MTC settings at every step and there are loads of printout. I included an status panel for ECA so that this is monitored in the SNO+ window. Also, the MTC GUI is kept in sync, so one can go there to see what are the current settings.

Source/Experiments/SNOP/LivePedestals.m
tlatorre-uchicago: How does it help to have pedestals running during PCA and N16 runs?

My concern is just having things centralized. In the future if we want to add something else to automatically run with embedded pedestals, ping crates, nhit monitor, etc, it will probably be easier to see everything happening in one place.

The current solution I have implemented in the ping crates pull request does so with a lock, which it is not necessary to be actually running in the same place for, so this would work with that, but I really don't like the lock solution, and I think if things get a little more complicated we're going to have to design some centralized class to make sure all of these things run without stepping on each other's toes.

Source/Objects/Custom Hardware/SNO+/MTC/ORMTCModel.m
tlatorre-uchicago: How many steps are there? I really think it's good to have these printouts as there is no other log of any of these things changing during the run. I think at the very least we need to have them printed out when the GUI is clicked.

tlatorre-uchicago:
For the embedded pedestals, how critical is the timing? IIRC these are only going to be used for charge pedestals is that right? One much more efficient way of getting a single pedestal instead of setting the pulser frequency and sleeping is to use the soft gt initiated pedestals. The only problem with these is that according to the documentation, 5% of the time the GT will come 20ns early. But if these are not being used for timing calibrations this would be a lot easier way to ensure that you get just one trigger.

Source/Experiments/SNOP/LivePedestals.m
jcaravaca: The N16 runs can be taken at different trigger rates (SNO had two different modes, low and high rates) than the physics or ECA rates. That means that potentially, we might have to introduce a correction to the default ECAs in calibration runs as well. That's why the embedded pedestals are used for. Same for PCA.

I think the approach I adopt here should work for every element: when you start your process (call it NHitMonitor, embedded pedestals or whatever) you set a flag in the object to 'Running' and you just need to check that that is False to allow to run the new process. We'll need to define a hierarchy for this, e.g. ECA - Ping Crates - NHitMonitor - Pedestals. We can discuss this in the next call.

Source/Objects/Custom Hardware/SNO+/MTC/ORMTCModel.m
jcaravaca: The embedded pedestals run in an infinite loop. THere are 32 steps that keep repeating, so the operator will be spammed every second with MTC logs.
I agree about printing it when we click in the GUI, so I'll revert this changes and we can work on reducing the printout in another PR.

jcaravaca:
We need this for timing calibrations. In fact, that's the first reason we want the embedded pedestals for since the low energy fitters work only with the timing information. I thought about what you say and it'd be way more elegant, but due to that HW bug we can't implement it that way.

Source/Experiments/SNOP/LivePedestals.m
tlatorre-uchicago: Are you talking about the time since last hit correction? If so, isn't the solution to just take ECAs at different rates? What does the trigger rate of the N16 or PCA have to do with it? In other words, what extra information do you get by taking ECAs within one of these runs?

Are you saying you think each of these processes should be responsible for checking all of the other ones?

tlatorre-uchicago:
Hm. I must be missing something. I see in the code you always use the same delay so this can't be used for time slope corrections correct? In what way is the timing information here used in calibrations? Is it used to just shift the whole timeslope curve based on this one point?

Source/Experiments/SNOP/LivePedestals.m
jcaravaca: Yes. There are 2 things that affect to the ECA calibration: TSLH and the FEC activity. We can take ECA at different rates but that'd take a long time and also it wouldn't correct by punctual situations like higher FEC activity periods. Embedded pedestals monitor in almost real life the actual state of the pedestal and allows to include a correction as a function of the absolute event time, rather than other parameters. Since the TSLH depend on the trigger rate, this needs to be monitored during the N16 and PCA runs as well. This was the intention behind implementing embedded pedestals.

You can have the processes lower in the hierarchy to check the other ones (as the NHitMonitor checks that we are not running ECAs or the live peds checks that we are not running the NHitMontor nor ECAs).

jcaravaca:
This has been tested OK at the teststand.

This is a summary of the new implementation of embedded pedestals (EP):
- EP are only allowed to run if the EP bit in the run type word is enabled.
- NHitMonitor is not allowed to run if the EP bit is enabled. This should be like this until we come up with a better solution.
- ECAs and PingCrates do not run either if the EP bit is checked
- If the EP bit is checked, the LivePed thread will be launched at run start. This will:
-- Unlock a radio button in the Calibration tab to turn ON or OFF the EP
-- The thread is always running in the background and checking if the radio button is ON, so that it can continue firing pedestals
-- When the EP are turned ON, the MTC settings are changed to fire pedestals at 1Hz, channel by channel
-- When the EP are turned OFF, the previous MTC settings are recovered
-- When the run stops/restarts, the livePed turns OFF the pedestals and dies

Also, I have included a 'Routine' panel to easily see if ECA/NHitMonitor/EP are running.

In the future I would like to better handle the way ECA/PingCrates/NHitMonitor/LivePeds run, to be able to run them simultaneously. However, I'd like to merge this as soon as possible so that we can take some long EP runs to understand if they are good for physics.

Source/Experiments/SNOP/LivePedestals.m
```diff
@@ -0,0 +1,338 @@
...
+
+- (bool) isNHitMonitorRunning
+{
+    bool nHitMonitorRunning = [[aSNOPModel nhitMonitor] isRunning];
+    return nHitMonitorRunning;
+}
+
+- (bool) shouldIrun
+{
+    bool nHitMonitorRunning = [self isNHitMonitorRunning];
```
tlatorre-uchicago: I think a better way of making sure that the nhit monitor and the embedded pedestals don't run at the same time is to use the lock that I've already created for this purpose. I already use it to avoid conflicts between ping crates and the nhit monitor. For an example see here https://github.com/snoplus/orca/blob/a7a34eb6c43ae79195d644d12e8b7b3317c68f3b/Source/Experiments/SNOP/SNOPModel.m#L1323. Basically you wrap the entire function which sends pedestals in another function which tries to acquire a lock first.

Source/Experiments/SNOP/LivePedestals.m
```diff
@@ -0,0 +1,338 @@
...
+                    [anMTCModel loadFineDelayToHardware];
+                    [anMTCModel loadPedestalCrateMaskToHardware];
+                    [anMTCModel loadPulserRateToHardware];
+                    [anMTCModel fireMTCPedestalsFixedRate];
+                });
+#endif
+
+            }
+
+            if([self isNHitMonitorRunning]){
```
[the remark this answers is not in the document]
jcaravaca: it's part of the 3 conditions in the loop:
- if( should I run ): run pedestals
- *else* if ( nhitmonitor is running ): send notification to update GUI status
- else: send notification to update GUI status

I was missing the first *else*

Source/Experiments/SNOP/LivePedestals.m
```diff
@@ -0,0 +1,338 @@
...
+
+}
+
+- (void) changePedestalMask:(NSMutableArray*)pedestal_mask
+{
+
+    //Set pedestal mask in FEC: this does not load it yet
+    for (ORFec32Model *fec in aFECModel) {
+        int crate_number = [fec crateNumber];
+        if(crate_number == 19) crate_number = 0; //hack for the teststand
```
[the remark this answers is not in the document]
jcaravaca: yeah, it's possible to be tested like this. The problem is that the ECA patterns do not exists for crate 19, so I need to consider crate 19 is crate 0...

Source/Experiments/SNOP/NHitMonitor.m
```diff
@@ -286,6 +286,7 @@ - (void) start: (int) crate pulserRate: (int) pulserRate numPulses: (int) numPul
     [runningThread start];
 
     [[NSNotificationCenter defaultCenter] postNotificationOnMainThreadWithName:ORNhitMonitorNotification object:self userInfo:nil];
+    [[NSNotificationCenter defaultCenter] postNotificationOnMainThreadWithName:ORRoutineChangedNotification object:self userInfo:@{@"routine":@"NHitMonitor"}];
```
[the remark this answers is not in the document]
jcaravaca: This notification is received by routineStatusChanged, which will call the [SNOP nhitmonitor] method to update the GUI status.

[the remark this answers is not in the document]
tlatorre-uchicago: Can you point to the line in the routineStatusChanged method which does this? I can't seem to find it.

Source/Experiments/SNOP/LivePedestals.m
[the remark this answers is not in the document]
tlatorre-uchicago: Ah, I see. Ok, I misunderstood that line.

Source/Experiments/SNOP/NHitMonitor.m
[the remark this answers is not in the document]
jcaravaca: Never-mind. I forgot I updated this code, which is now handled by the ORNhitMonitorNotification notification, so that we don't have duplicated notifications. I'll remove those lines.

jcaravaca:
The last commit should cover most of the comments.

The only tricky part is the ECAlock implementation: the embedded pedestals always launches the LivePed thread, which will periodically check if embedded pedestals can be ran. In this case the ECAlock needs to be checked periodically within the loop, so someone I need to know if the lock was set previously by LivePed itself or someone else... I'll give a thought to this and push a solution.

jcaravaca:
Tested at the TS.

The lock works really well:
- When the embedded pedestals are ongoing, the NHitMonitor waits 10 seconds to acquire the lock and viceversa
- If the NHitMonitor is ongoing the embedded pedestals show 'Waiting for lock' and they are resume as soon as the NHitMonitor is done
- Couldn't test with ping crates, but I assume the functionality is the same than the NHitMonitor and then it should work fine

tlatorre-uchicago:
There is one XCode warning that `prev_pedwidth` is not used anywere in LivePedestals.

jcaravaca:
I have added the suggested label and now the button is grayed out when the Embedded Ped. bit is not enabled. I have removed the extra variable as well. I tested it locally and seemed to work fine. I can give it a try tomorrow at the TS.

jcaravaca:
Tested OK at the TS.

tlatorre-uchicago:
merged.

PR #400, opened 2017-05-05
SNO+: Tidy redisClient calls in TUBii model

EdLeming:
There were a couple of instances where intCommand was called directly on a redisClient object without being wrapped in try catch. They were associated with getters for the sync/Async trigger masks, which get called in the ELLIE model at the start of an ELLIE flash sequence. This could well have been the source of the uncaught redisClient exceptions causing the recent beaching in ELLIE runs.


tlatorre-uchicago:
This looks good, but it raises a couple of issues:

1. The "Unexpected Response Type" message means that the TUBii server returned the wrong data type to a command.
2. It looks like for IntCmd, any exception is caught and then the user is returned a nil object. This means that every single call to one of these commands, it is necessary to check for that or alternatively to add a method which lets the exception propagate. Can you do a quick sweep of the ELLIE related code and make sure that that is the case?
3. For the method OkCmd there is no way to find out whether the command succeeded. Can you also do a sweep and make sure that you are not calling any OkCmd? If so, we should figure out a way to add a method which will let the exception propagate.

EdLeming:
@tlatorre-uchicago. Bringing this up with the TUBii guys offline.

EdLeming:
@tlatorre-uchicago. OK, so Eric and I have talked about this propagation of exceptions before. Eric even made this branch to propagate them through, but it looks like it never got merged in.

https://github.com/marzece/orca/tree/tubii_exceptions

Based on this branch I've removed the try catch from the [tubiiModel sendOkCmd/sendIntCmd] method so exceptions propagate. I've found a couple of tubiiModel calls in the ELLIE model which were not wrapped, and amended. All ELLIE model calls should now be safe. I've also wrapped all model calls in the TUBiiController. 

@marzece please have a quick look over.  Make sure i haven't done anything stupid.

tlatorre-uchicago:
@marzece let me know when you've gotten a chance to look at this.

tlatorre-uchicago:
Let me know when you get a chance to test this on the detector @edleming and I'll merge.

EdLeming:
👍  

@ecaden, I'm a little out of the loop on the detector schedule. Do we have any LV coming up?

EdLeming:
@tlatorre-uchicago, testing was successful. 

Unfortunately the beaching observed is a rare failure mode so can't be sure it has actually fixed the problem. Will need another 100+ run tellie shift or two until we can start feeling comfortable.

Even if not, this was a useful tidying exercise. The exception handing logic flow is now much more robust.

tlatorre-uchicago:
merged.

PR #402, opened 2017-05-11
SNO+: Remove intensity correlation from trigger rate check for TELLIE

EdLeming:
Recent testing of the PCA set points has shown that TELLIE's calibrated photon output doesn't correlate with the nHits observed by the detector. This PR replaces an intensity vs rate check (which uses those 'bad' calibrations!) with a simple rate cut at 1kHz.


tlatorre-uchicago:
merged.

PR #403, opened 2017-05-15
SNO+: initialize nhit monitor before loading settings

tlatorre-uchicago:
This pull request fixes a few bugs in the initialization of the nhit monitor settings.

- delete initialization of the nhit monitor in init() since initWithCoder() is immediately called afterwards even when the SNO+ controller is first added to the experiment
- initialize the nhit monitor object before loading settings in initWithCoder()
- don't automatically run the nhit monitor until all settings are loaded

This is an attempt to fix a bug that was seen on May 14, 2017 where it appeared that there were two copies of the nhit monitor running (i.e. two nstimer objects had somehow been started). None of these fixes obviously fixes the issue but since some of them may have started a timer with uninitialized values it's possible that it might fix the issue.


tlatorre-uchicago:
merged.

PR #405, opened 2017-05-17
SNO+: added autorelease pools to couchdb calls in ELLIEModel

EdLeming:
Crashes were observed during particularly long smellie runs: 100478 and 100469. 

After every sub run ORCA goes and updates the SMELLIE_RUN couchdb document, appending it with the information associated with that sub-run. Offline investigations have shown that calls to ORCouchDB were leaking memory. After approximately 700 sub-runs ORCA would run out of memory and die. Autorelease pools have been added inside the couch request functions which have brought the memory usage back under control. 

I believe this was not noticed before as a autorelease pools already exist around the smellie_run method in the ELLIEModel, which would have cleaned up the leaky memory once that function completed. We needed a particularly long sequence like this to flag the problem.


tlatorre-uchicago:
merged.

PR #406, opened 2017-05-23
ping crates at run starts

tlatorre-uchicago:
This commit updates the SNO+ model to ping each slot in the detector once at
the start of physics run. These events will be useful to diagnose any problems
with the triggers on a run by run basis.

@tannerbk is working on a nearline job that will analyze these events to look for issues and alert the operator if there is a problem.

This hasn't been tested yet at the teststand.

One issue that needs to be figured out is how to mark these events in the data stream. Currently, I use a standard EPED record to and use the half crate id field to mark which crate and slot are being pedestalled. I think the embedded pedestals also will use this record to mark which channels are being pedestalled, so we probably need to come up with a unique flag to identify these records. In addition, we'll need to figure out some way to synchronize the embedded pedestals so that they don't start until after this finishes running.

### DIFF
--- a/Source/Experiments/SNOP/NHitMonitor.h
+++ b/Source/Experiments/SNOP/NHitMonitor.h
@@ -50,6 +50,7 @@ struct NhitRecord {
 - (void) dealloc;
 - (void) registerNotificationObservers;
 - (void) runAboutToStop: (NSNotification*) aNote;
+- (void) orcaAboutToQuit: (NSNotification*) aNote;
 - (void) _waitForThreadToFinish;
 - (BOOL) isRunning;
 - (void) stop;
--- a/Source/Experiments/SNOP/NHitMonitor.h
+++ b/Source/Experiments/SNOP/NHitMonitor.h
@@ -58,7 +58,8 @@ struct NhitRecord {
 - (void) disconnect;
 - (void) nhitMonitorCallback: (ORPQResult *) result;
 - (void) run: (NSDictionary *) args;
-- (void) _run: (int) crate pulserRate: (int) pulserRate numPulses: (int) numPulses maxNhit: (int) maxNhit slots: (NSMutableArray *) slots channels:(NSMutableArray *) channels;
+- (void) _run: (NSDictionary *) args;
+- (void) __run: (int) crate pulserRate: (int) pulserRate numPulses: (int) numPulses maxNhit: (int) maxNhit slots: (NSMutableArray *) slots channels:(NSMutableArray *) channels;
 @end
 
 extern NSString* ORNhitMonitorUpdateNotification;
--- a/Source/Experiments/SNOP/NHitMonitor.m
+++ b/Source/Experiments/SNOP/NHitMonitor.m
@@ -17,6 +17,7 @@
 #import "ORGlobal.h"
 #import "RedisClient.h"
 #import "ORMTC_Constants.h"
+#import "TUBiiModel.h"
 
 #define SWAP_INT32(a,b) swap_int32((uint32_t *)(a),(b))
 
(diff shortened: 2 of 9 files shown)


jcaravaca:
Currently the 'calType' in the EPED records are used to mark which kind of pedestals we are getting. For the ECAs I do:
- Pedestal type = (fCalType / 10) % 10 (1->PDST, 2-> TSLP, 3->LivePeds, 4->NHitMonitor)
- Pattern = fCalType % 10 (For ECAs -> pedestal masks, for LivePeds -> Channel number)
You could use:
- Pedestal type = 5
- Pattern = half-crate ID

tlatorre-uchicago:
Updated this to use 50 for the calType. @jcaravaca does that look OK?

In order to synchronize pedestals between the nhit monitor, ping crates, and the embedded pedestals I added a NSLock object called ecaLock to the SNO+ model. The idea is that before performing any pedestal related hardware changes, each function tries to get a hold of the lock and only performs those changes while it has control of the lock. This is not the best implementation, because ideally you would like to have things run serially and not always have to poll locks, but I think it will work.

Currently what the ping crates and nhit monitor do is to try to acquire the lock for 10 seconds. If they can't get it after 10 seconds, then they quit. If they do get a hold of the lock they run and then release the lock at the end.

It's absolutely critical that the lock gets released at the end, so what I've done is to create a new function which wraps the function doing all the work and put a try finally block to make sure that the lock is released no matter what. Note that it is absolutely critical to release the lock in the same thread as it was locked in or it will result in undefined behavior. Basically, what I did was the following:

```
- (void) pingCratesAtRunStart
{
    /* This function is called on the main thread and launches another thread to ping the crates. */
    [NSThread detachNewThreadSelector:@selector(_pingCratesAtRunStart)
                             toTarget:self
                           withObject:nil];
}

- (void) _pingCratesAtRunStart
{
    /* This function runs in a separate thread and is mostly just a wrapper for __pingCratesAtRunStart.
     * The reason is that we need to acquire the lock first and make sure we release it even if an 
     * exception gets thrown. */
    @autoreleasepool {
        /* Try to acquire the lock for 10 seconds. If we can't get it then we
         * just skip the ping crates. */
        if ([ecaLock lockBeforeDate:[NSDate dateWithTimeIntervalSinceNow:10.0]]) {
            @try {
                [self __pingCratesAtRunStart];
            } @finally {
                /* Make sure to unlock the lock when we're done. */
                [ecaLock unlock];
            }
        } else {
            NSLogColor([NSColor redColor],
                       @"pingCratesAtRunStart: unable to acquire eca lock!\n");
        }
    }
}

- (void) __pingCratesAtRunStart
{
    /* This function is where we actually ping the crates. */
}
```

Another thing to note is that it's critical that all of the pedestal related settings are set back to what they were before the thread runs. This is currently handled in __pingCratesAtRunStart.

tlatorre-uchicago:
Tested on the teststand on May 28, 2017. Then tested on the detector on the same day. The two runs which have the ping crates at the run start are 101079 and 101080.

tannerbk:
As per our discussion it would be nice if this set a different pattern in the header at the end of the ping.

tannerbk:
It would be really nice if we could get this into Orca with all the recent issues with the trigger signals.

tlatorre-uchicago:
This is currently waiting on the pedestal test. One alternative is to do something similar to the check rates script and set EXT2 high while it's running.

tlatorre-uchicago:
I updated the pull request so that it sets the EXT2 high when the ping crates run. Tested on the detector on June 27, 2017.

- verified that the ping crates actually runs
- verified that the eca lock works by starting the nhit monitor before the ping crates and verifying that the ping crates quits since it can't acquire the lock and vice versa
- verified that EXT2 bit is set for all events while the ping crates is running

tlatorre-uchicago:
merged.

PR #410, opened 2017-05-25
add method to enable the extended readout mapping on the CAEN board

tlatorre-uchicago:
This commit enables the undocumented extended readout mapping for the CAEN board which speeds up the readout.

This shouldn't be merged until the new mtc server is installed.


tlatorre-uchicago:
Tested on the teststand on May 26, 2017.

tlatorre-uchicago:
merged.

PR #411, opened 2017-05-26
Phil/abort run stop

boardhead:
Add ability to force a run to stop while waiting for the data buffers to empty.  Dialog currently appears after 3 seconds of waiting -- we may want to change this.


tlatorre-uchicago:
merged.

PR #412, opened 2017-05-27
Add TUBii and CAEN settings in Standard Runs

jcaravaca:
Add TUBii and CAEN settings in Standard Runs and display in the SNO+ panel the current and DB values. I changed the functionality in TUBii to store in the CurrentState structure the state of the model, not the HW, since that's the way eveything else works in ORCA. Since that structure wasn't used, I don't think it should be a problem, but it'd need approval from @marzece and @icoulter.

Another issue is that the CurrentState is kept in sync with the SNO+ and TUBii GUI, but currently to update the model you need to hit 'Send to TUBii'. In order to update the model when anything changes in the TUBii GUI I need to include actions for every single element. So I'd like to reach a consensus before I implement that.

I include a snapshot of the new SNO+ control panel with the elements I'm storing, in case I'm missing something. I've added a new 'Calibrations' tab which will host ELLIE and ECAs (@EdLeming might want to comment). This PR needs to be merged AFTER the supernova_ECA one to not mess up much with the GUI changes.

image


Source/Objects/Custom Hardware/SNO+/CAEN/SNOCaenModel.m
```diff
@@ -415,6 +416,7 @@ - (void) setEnabledMask:(unsigned short)aEnabledMask
     enabledMask = aEnabledMask;
 	
     [[NSNotificationCenter defaultCenter] postNotificationName:SNOCaenModelEnabledMaskChanged object:self];
+    [[NSNotificationCenter defaultCenter] postNotificationOnMainThreadWithName:SNOCaenSettingsChanged object:self];
```
tlatorre-uchicago: I think it would be more efficient to add observers for each of these notifications in the SNO+ model rather than post two notifications every time the CAEN settings change.

Source/Objects/Custom Hardware/SNO+/CAEN/SNOCaenModel.m
```diff
@@ -1252,6 +1274,65 @@ - (void) encodeWithCoder:(NSCoder*) anEncoder
...
+        [CAENStateDict setObject:[NSNumber numberWithUnsignedShort:[self coincidenceLevel]] forKey:[self getStandardRunKeyForField:@"coincidenceLevel" ]];
+        [CAENStateDict setObject:[NSNumber numberWithUnsignedLong:[self triggerOutMask]] forKey:[self getStandardRunKeyForField:@"triggerOutMask" ]];
+        [CAENStateDict setObject:[NSNumber numberWithUnsignedLong:[self postTriggerSetting]] forKey:[self getStandardRunKeyForField:@"postTriggerSetting" ]];
+        [CAENStateDict setObject:[NSNumber numberWithUnsignedLong:[self frontPanelControlMask]] forKey:[self getStandardRunKeyForField:@"frontPanelControlMask" ]];
+        [CAENStateDict setObject:[NSNumber numberWithUnsignedShort:[self enabledMask]] forKey:[self getStandardRunKeyForField:@"enabledMask" ]];
+        for (int idac=0; idac<kNumCaenChannelDacs; ++idac) {
+            [CAENStateDict setObject:[NSNumber numberWithUnsignedShort:[self dac:idac]] forKey:[NSString stringWithFormat:@"%@_%i",[self getStandardRunKeyForField:@"dac"],idac]];
+        }
+    } @catch(...){
+        NSLogColor([NSColor redColor], @"CAEN: settings couldn't not be saved \n");
```
tlatorre-uchicago: If you're going to catch the exception here, you need to relay that info back to the caller so it doesn't save a dictionary with invalid values. There are two ways to do this:

1. don't catch the exception. Then you need to make sure any upstream code has exception handlers.
2. Pass back NULL instead of an NSDictionary pointer, and then make sure all calls to this function check the return value.

I think the best way depends on how often something like this will happen. Since these are just getting objects from the model it's very unlikely to fail, and so I would just let the exception bubble up and put try catch blocks around the code which gets the current state.

Source/Objects/Custom Hardware/SNO+/CAEN/SNOCaenModel.m
```diff
@@ -1252,6 +1274,65 @@ - (void) encodeWithCoder:(NSCoder*) anEncoder
...
+        [self setCoincidenceLevel:[[settingsDict objectForKey:[self getStandardRunKeyForField:@"coincidenceLevel"]] unsignedShortValue]];
+        [self setTriggerOutMask:[[settingsDict objectForKey:[self getStandardRunKeyForField:@"triggerOutMask"]] unsignedLongValue]];
+        [self setPostTriggerSetting:[[settingsDict objectForKey:[self getStandardRunKeyForField:@"postTriggerSetting"]] unsignedLongValue]];
+        [self setFrontPanelControlMask:[[settingsDict objectForKey:[self getStandardRunKeyForField:@"frontPanelControlMask"]] unsignedLongValue]];
+        [self setEnabledMask:[[settingsDict objectForKey:[self getStandardRunKeyForField:@"enabledMask"]] unsignedShortValue]];
+        for (int idac=0; idac<kNumCaenChannelDacs; ++idac) {
+            [self setDac:idac withValue:[[settingsDict objectForKey:[NSString stringWithFormat:@"%@_%i",[self getStandardRunKeyForField:@"dac"],idac]] unsignedShortValue]];
+        }
+    } @catch(...){
+        NSLogColor([NSColor redColor], @"CAEN: settings couldn't not be loaded \n");
```
tlatorre-uchicago: Same here.

Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiController.m
```diff
@@ -7,7 +7,9 @@
 //
 
 #import "TUBiiController.h"
+#pragma pack(push)
```
tlatorre-uchicago: What is this here for?

Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiController.m
```diff
@@ -186,6 +193,71 @@ - (void) tubiiLockChanged:(NSNotification*)aNotification
...
+    struct TUBiiState theTUBiiState = [model currentState];
+    NSUInteger trigMaskVal = (theTUBiiState.syncTrigMask | theTUBiiState. asyncTrigMask);
+    NSUInteger syncMaskVal = 16777215 - theTUBiiState.asyncTrigMask;
+    [self SendBitInfo:trigMaskVal FromBit:0 ToBit:24 ToCheckBoxes:TrigMaskSelect];
+    [self SendBitInfo:syncMaskVal FromBit:24 ToBit:48 ToCheckBoxes:TrigMaskSelect];
+    //CAEN
+    CAEN_CHANNEL_MASK ChannelMask = theTUBiiState.CaenChannelMask;
+    CAEN_GAIN_MASK GainMask = theTUBiiState.CaenGainMask;
+    BOOL err = YES;
+    err &= [caenChannelSelect_0 selectCellWithTag:(ChannelMask & channelSel_0)>0];
```
tlatorre-uchicago: I'm not sure I understand this code. err is a BOOL which defaults to YES and you are bitwise anding it with the results of these functions?

Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiModel.m
```diff
@@ -724,38 +831,63 @@ - (void) setCurrentState:(struct TUBiiState)aState {
...
+        [TubiiStateDict setObject:[NSNumber numberWithUnsignedInt:currentState.CaenGainMask] forKey:[self getStandardRunKeyForField:@"CaenGainMask"]];
+        [TubiiStateDict setObject:[NSNumber numberWithUnsignedInt:currentState.speakerMask] forKey:[self getStandardRunKeyForField:@"speakerMask"]];
+        [TubiiStateDict setObject:[NSNumber numberWithUnsignedInt:currentState.counterMask] forKey:[self getStandardRunKeyForField:@"counterMask"]];
+        [TubiiStateDict setObject:[NSNumber numberWithUnsignedInteger:currentState.MTCAMimic1_ThresholdInBits] forKey:[self getStandardRunKeyForField:@"MTCAMimic1_ThresholdInBits"]];
+        [TubiiStateDict setObject:[NSNumber numberWithUnsignedInt:currentState.DGT_Bits] forKey:[self getStandardRunKeyForField:@"DGT_Bits"]];
+        [TubiiStateDict setObject:[NSNumber numberWithUnsignedInt:currentState.LO_Bits] forKey:[self getStandardRunKeyForField:@"LO_Bits"]];
+        [TubiiStateDict setObject:[NSNumber numberWithUnsignedInt:currentState.controlReg] forKey:[self getStandardRunKeyForField:@"controlReg"]];
+    } @catch(...){
+        NSLogColor([NSColor redColor], @"TUBii: settings couldn't not be saved \n");
```
tlatorre-uchicago: Same issue as mentioned above.

Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiModel.m
```diff
@@ -724,38 +831,63 @@ - (void) setCurrentState:(struct TUBiiState)aState {
...
+        currentState.CaenChannelMask = [[settingsDict objectForKey:[self getStandardRunKeyForField:@"CaenChannelMask"]] floatValue];
+        currentState.CaenGainMask = [[settingsDict objectForKey:[self getStandardRunKeyForField:@"CaenGainMask"]] floatValue];
+        currentState.counterMask = [[settingsDict objectForKey:[self getStandardRunKeyForField:@"counterMask"]] floatValue];
+        currentState.speakerMask = [[settingsDict objectForKey:[self getStandardRunKeyForField:@"speakerMask"]] floatValue];
+        currentState.MTCAMimic1_ThresholdInBits = [[settingsDict objectForKey:[self getStandardRunKeyForField:@"MTCAMimic1_ThresholdInBits"]] floatValue];
+        currentState.DGT_Bits = [[settingsDict objectForKey:[self getStandardRunKeyForField:@"DGT_Bits"]] floatValue];
+        currentState.LO_Bits = [[settingsDict objectForKey:[self getStandardRunKeyForField:@"LO_Bits"]] floatValue];
+        currentState.controlReg = [[settingsDict objectForKey:[self getStandardRunKeyForField:@"controlReg"]] floatValue];
+    } @catch(...){
+        NSLogColor([NSColor redColor], @"TUBii: settings couldn't not be loaded \n");
```
tlatorre-uchicago: Same issue as mentioned above.

tlatorre-uchicago:
Overall, I think it looks good. I made a few comments about the pull request. I like the idea of updating the TUBii model to store the state of the model and not the hardware so that it is the same as the other objects, but I don't think we should have two copies of every single model variable. You could update it be like the CAEN model and keep the model variables as class variables and only serialize in the currentStateToDict method or you could update everything to use the struct. @marzece @icoulter any opinions?

Source/Objects/Custom Hardware/SNO+/CAEN/SNOCaenModel.m
jcaravaca: I thought it'd be more compact this than catching all the notifications, which are quite a lot. I'm fine with either.

Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiController.m
jcaravaca: Remnant of early testing. I'll remove it.

jcaravaca:
This is mostly ready to go. Few caveats:
- There are currently two bugs in TUBii that I'm not sure how to solve:
-- When you set a trigger mask (sync and async) and try to recover it, the values are different. This is due to some bitwise logic I didn't spend time reviewing, but it'd be nice if some expert crosscheck this. I just copied and paste the logic, but I don't discard I might have missed something somewhere.
-- Currently, the 'Counter Advanced Options' set bits in the control register, but do not unset them properly. I didn't want to change anything of the logic just in case, but someone please review that.
- I have changed the TUBii controller quite a lot, so we'll need to go through normal operation to test everything makes sense
- I rearranged the GUI. Now I show ALL the DB values along with the current values in the SR tab. ECA and ELLIE has been moved to the calibrations tab (see below).
- The Lockout width has been included in the SR definition as suggested in #421 

image

image

icoulter:
Hi Javi,
I'll take a look at the trigger logic tomorrow. Sorry, I meant to today but I forgot. I didn't notice this last time but its possible I didn't check for it.

tannerbk:
It might be worth considering what CAEN and TUBII settings actually need to be displayed in the standard run panel. I think a lot of those settings are never touched, so it might not be worth crowding the display with their values.

Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiController.m
```diff
@@ -938,7 +1052,7 @@ - (void) SendBitInfo:(NSUInteger) maskVal FromBit:(int)low ToBit:(int) high ToCh
     //Helper function to send a bit value to a bunch of check boxes
     for (int i=low;i<high;i++)
     {
-        if((maskVal & 1<<(i-low)) >0)
+        if( (maskVal & 1<<i)>0 )
```
icoulter: Any reason this line was changed? It seems to be the cause of most of this trigger mask strangeness?

Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiController.m
```diff
@@ -922,6 +908,134 @@ - (IBAction)ClockSourceMatchHardware:(id)sender {
     }
 }
 
+- (IBAction)trigMaskAction:(id)sender {
+    NSUInteger trigMaskVal = [self GetBitInfoFromCheckBoxes:sender FromBit:0 ToBit:24];
+    NSUInteger syncMaskVal = [self GetBitInfoFromCheckBoxes:sender FromBit:24 ToBit:48];
+
+    NSLogColor([NSColor redColor], @"trigMaskAction trigMaskVal: 0x%X\n",trigMaskVal);
+    NSLogColor([NSColor redColor], @"trigMaskAction syncMaskVal: 0x%X\n",syncMaskVal);
```
icoulter: Remember to remove these debugging comments before this is merged.

Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiController.m
```diff
@@ -922,6 +908,134 @@ - (IBAction)ClockSourceMatchHardware:(id)sender {
...
+            if(trigMaskVal & (1<<i))
+                asyncMask |= 1<<i;
+            else
+                asyncMask &= ~(1<<i);
+            syncMask &= ~(1<<i);
+        }
+    }
+
+    NSLogColor([NSColor redColor], @"trigMaskAction syncMask: 0x%X\n",syncMask);
+    NSLogColor([NSColor redColor], @"trigMaskAction asyncMask: 0x%X\n",asyncMask);
```
icoulter: Same as above.

icoulter:
Speaker and counter masks are still acting up. Do you know why the tags of the some of the check boxes in the .xib changed? That's probably related. The MTCA Mimics have changed values as have ELLIE and the GT.

Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiController.m
jcaravaca: I changed this function to make the speaker and counter check boxes work.
I assumed this wants to set the provided mask from only bits low to high, and previously it was setting the bits from 0 to bit high-low. Am I misunderstanding something here?

jcaravaca:
@icoulter I reverted the changes on the tags that were scrambled for some reason.
That doesn't solve the problem with the speaker/counter masks when reverting the functionality of the SendBitInfo function, though.

Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiController.m
jcaravaca: If I revert this change, the trigger masks show still a weird behaviour.
Are the sync/async checkboxes suppose to be exclusive? I mean, only one of the check boxes can be enabled for each bit?

icoulter:
The tags weren't scrambled. The buttons were added to the gui in the wrong order due to a miscommunication between Eric and I and the tags were changed to compensate (as the user doesn't actually care what order they are in as long as they are labelled correctly). They shouldn't be changed.
You are right it doesn't solve the greater issue though.

Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiController.m
icoulter: Ok, that'll break the trigger mask though. If I recall, I compensated for this in the counters by shifting the bits.

In your way you are sending the low'th bit of the mask to checkbox low .
Mine I am sending the 0th bit of the mask to checkbox low.
I found the second an easier way to visualise things. Especially in the trigger mask were bits 24-48 represent a completely separate word so I either need to do this or shift it by 24 bits before sending it.

icoulter: I don't follow your meaning.
Due to an issue with how I implemented it in the hardware, the sync/async doesn't have a value (or physical meaning) unless that bit is set in the trigger mask. Is that what you mean? In your implementation, it makes them default to sync.

jcaravaca:
After some back and forth @icoulter and I figured out the problem, so this PR should be ready for review. I haven't test it yet at the detector.

icoulter:
As Javi says, we've discussed this and I've did some tests here. I think we've worked out the issues and it works as expected. Next step is to test it in the detector.

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -647,6 +649,12 @@ - (void) runInitialization:(NSNotification*)aNote
             goto err;
         }
 
+        /* Load the TUBii settings to hardware. */
+        if ([tubii sendCurrentStateToHW]) {
```
tlatorre-uchicago: The tubii pointer here is never set to anything.

Source/Objects/Custom Hardware/SNO+/CAEN/SNOCaenModel.m
```diff
@@ -1271,6 +1293,67 @@ - (void) encodeWithCoder:(NSCoder*) anEncoder
     }
 }
 
+- (NSDictionary*) CurrentStateToDict
```
tlatorre-uchicago: We should probably keep these methods named the same across different objects. Currently, it's called `serializeToDictionary` in the MTC model. Can you rename this and the TUBii method to that?

Source/Objects/Custom Hardware/SNO+/CAEN/SNOCaenModel.m
```diff
@@ -1271,6 +1293,67 @@ - (void) encodeWithCoder:(NSCoder*) anEncoder
...
+{
+
+    //Dump CAEN settings in dictionary
```
tlatorre-uchicago: Worth adding a comment that this returns NULL if there was an error or removing the try catch block in case it fails so that the caller can catch any errors.

Source/Objects/Custom Hardware/SNO+/CAEN/SNOCaenModel.m
```diff
@@ -1271,6 +1293,67 @@ - (void) encodeWithCoder:(NSCoder*) anEncoder
...
+        [CAENStateDict setObject:[NSNumber numberWithUnsignedLong:[self frontPanelControlMask]] forKey:[self getStandardRunKeyForField:@"frontPanelControlMask" ]];
+        [CAENStateDict setObject:[NSNumber numberWithUnsignedShort:[self enabledMask]] forKey:[self getStandardRunKeyForField:@"enabledMask" ]];
+        for (int idac=0; idac<kNumCaenChannelDacs; ++idac) {
+            [CAENStateDict setObject:[NSNumber numberWithUnsignedShort:[self dac:idac]] forKey:[NSString stringWithFormat:@"%@_%i",[self getStandardRunKeyForField:@"dac"],idac]];
+        }
+
+        return CAENStateDict;
+
+    } @catch(...){
+        NSLogColor([NSColor redColor], @"CAEN: settings couldn't not be saved \n");
```
tlatorre-uchicago: Should add here what the error is so that we can debug it if it happens. For example,

```objective-c
NSLogColor([NSColor redColor], @"CAEN: settings couldn't be saved. error: %@ reason: %@", [e name], [e reason]);
```

Source/Objects/Custom Hardware/SNO+/CAEN/SNOCaenModel.m
```diff
@@ -1271,6 +1293,67 @@ - (void) encodeWithCoder:(NSCoder*) anEncoder
...
+        return NULL;
+    }
+
+}
+
+- (void) setCurrentStateFromDict:(NSMutableDictionary*)settingsDict {
```
tlatorre-uchicago: Same as above, should be called `LoadFromSerialization`.

Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiModel.m
```diff
@@ -688,60 +835,79 @@ - (BOOL) CounterMode {
...
+        [TubiiStateDict setObject:[NSNumber numberWithUnsignedInt:currentState.speakerMask] forKey:[self getStandardRunKeyForField:@"speakerMask"]];
+        [TubiiStateDict setObject:[NSNumber numberWithUnsignedInt:currentState.counterMask] forKey:[self getStandardRunKeyForField:@"counterMask"]];
+        [TubiiStateDict setObject:[NSNumber numberWithUnsignedInteger:currentState.MTCAMimic1_ThresholdInBits] forKey:[self getStandardRunKeyForField:@"MTCAMimic1_ThresholdInBits"]];
+        [TubiiStateDict setObject:[NSNumber numberWithUnsignedInt:currentState.DGT_Bits] forKey:[self getStandardRunKeyForField:@"DGT_Bits"]];
+        [TubiiStateDict setObject:[NSNumber numberWithUnsignedInt:currentState.LO_Bits] forKey:[self getStandardRunKeyForField:@"LO_Bits"]];
+        [TubiiStateDict setObject:[NSNumber numberWithUnsignedInt:currentState.controlReg] forKey:[self getStandardRunKeyForField:@"controlReg"]];
+        return TubiiStateDict;
+    } @catch(...){
+        NSLogColor([NSColor redColor], @"TUBii: settings couldn't not be saved \n");
```
tlatorre-uchicago: Same as above, should print out the error if it occurs here or get rid of the try catch block.

Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiModel.m
```diff
@@ -688,60 +835,79 @@ - (BOOL) CounterMode {
...
+        [self setTrigMask: currentState.syncTrigMask setAsyncMask:currentState.asyncTrigMask];
+        [self setCaenMasks:currentState.CaenChannelMask GainMask:currentState.CaenGainMask];
+        [self setSpeakerMask: currentState.speakerMask];
+        [self setCounterMask: currentState.counterMask];
+        [self setMTCAMimic1_ThresholdInBits: currentState.MTCAMimic1_ThresholdInBits];
+        [self setGTDelaysBits:currentState.DGT_Bits LOBits:currentState.LO_Bits];
+        [self setControlReg: currentState.controlReg];
+    }
+    @catch(...){
+        return 0;
```
tlatorre-uchicago: Same as above, should print out the error here or remove the try-catch block.

jcaravaca:
Implemented Tony's comments.

jcaravaca:
Tested on site and fixed 2 minor bugs related to a wrong conversion from bits to ns. This should be ready to go now.

jcaravaca:
The only change in the GUI since last time is the renaming of two run type bits. That was merged fine and the GUI works properly. As I said in the call, there are new buttons for future functionality regarding ECAs and embedded pedestals. Those are currently grayed out (see pics above). In this way, we can work on GUI changes while we merge the 2 pendent PRs.

tlatorre-uchicago:
There's one memory leak and a couple of previous comments that weren't addressed. Once these are fixed let me know and I'll merge.

If you can send me an email when you get finished migrating the Version 1 -> Version 2 for the standard runs I will deploy it on the monitoring machines. It might be nice while doing so to go ahead and update the CAEN settings for the SMELLIE runs so we don't have to burn any detector time changing those settings.

EdLeming:
I second Tony's SMELLIE comment. The SMELLIE specific tubii settings should be added at the same time. Let me know if I should send on a list of those settings to anyone.

tlatorre-uchicago:
merged.

PR #415, opened 2017-06-01
	Change delay to 10 seconds

boardhead:
Dialog now waits 10 seconds before giving the option to force a run stop


tlatorre-uchicago:
merged.

PR #417, opened 2017-06-04
SNO+: Update how fibre field is passed in TELLIE expert tab

EdLeming:
In the TELLIE expert tab the fibre name was being read from the GUI's drop down menu. This caused a bug: If a user manually changes the TELLIE channel to be fired (rather than using the usual fibre / node selection), the fibre field is automatically cleared. In this case the fibre field would be left blank and no fibre information is propagated to the couchdb during running. This PR fixes this bug.


tlatorre-uchicago:
merged.

PR #422, opened 2017-06-14
Add the SetTellieMode command to the TUBiiModel.

icoulter:
@EdLeming As promised.
I've added it to the TUBii Model.
Haven't added it to the controller as adding it to the xib will conflict with #412. That's probably less essential anyway as the Tellie pulser typically isn't run by hand from the TUBii window.


tlatorre-uchicago:
merged.

PR #423, opened 2017-06-16
SNO+: Allow running non-defaults to non-experts.

jcaravaca:


tlatorre-uchicago:
merged.

PR #424, opened 2017-06-16
Nominal settings check hv

tlatorre-uchicago:
This pull request adds a check to make sure that any channels with HV on have their sequencer enabled to avoid blind flashers.

Tested on the teststand on June 16, 2017.


tlatorre-uchicago:
merged.

PR #427, opened 2017-06-21
Add TUBii's PGT into the lock.

icoulter:
Noticed that TUBii's PGT setting weren't included in the physics lockout. This should fix that.


tlatorre-uchicago:
merged.

PR #428, opened 2017-06-27
zero all pedestal masks before physics runs

tlatorre-uchicago:
This pull request sets the pedestal mask for all FECs to zero before a physics run. The reason for this is that it was noticed that the number of channels in the pedestal mask seems to affect the noise on the trigger signal. See this shift report for more details: http://snopl.us/shift/view/9e1ff17e58704756a99f947ec2509f39?index_start=15.


tlatorre-uchicago:
merged.

PR #433, opened 2017-06-29
set calType to zero when ping crates at run start finishes

tlatorre-uchicago:
This pull request updates the ping crates at run start method to set the calType to zero when it finishes. The reason is that RAT associates all events after an EPED record with that EPED record, and so this makes it easier for nearline jobs to just look at the ping crates events.


tlatorre-uchicago:
merged.

PR #435, opened 2017-07-27
add a run type bit for AV recirculation

tlatorre-uchicago:


tlatorre-uchicago:
merged. This doesn't look particularly nice since the words get cut off in the GUI, but I don't want to change much in the GUI since there are open pull requests which have modified the SNO+ model GUI.

PR #437, opened 2017-08-09
SNO+: Properly handle channel number from DB

EdLeming:
This is a one line fix for a bug introduced in my last PR. TELLIE channel numbers are strings in the database document, not ints.


EdLeming:
ah, yeah. I had to do something to fix this when testing UG on the other branch. Conflict now resolved.

tlatorre-uchicago:
merged.

PR #438, opened 2017-08-14
fix a few warnings pointed out by the XCode analyzer tool

tlatorre-uchicago:
I ran the XCode analyzer tool in XCode 6.2 and it pointed out a couple of unused variables.


tlatorre-uchicago:
merged.

PR #440, opened 2017-08-15
Unlock GUI settings if in expert mode

jcaravaca:
Unlock GUI settings if in physics run. This PR allows to change SR settings during any type of running, yet preventing from loading settings into HW if not in Maintenance.
The change in the CAEN GUI is just moving the label 'In maintenance' closer to the 'Init board' button.


tlatorre-uchicago:
Sweet! The Set button in the Front Panel IO Control box in the CAEN window should also probably be greyed out.

tlatorre-uchicago:
Is it also possible to unlock the run type word? With that change we could easily edit the standard run settings.

I guess it depends on when it gets latched into the Orca run word, i.e. what would happen if you changed the run type checkboxes and then the run rolled over or got restarted? If it would automatically reset the run type word then we should be fine.

jcaravaca:
Currently unlocking the RunTypeWord results in shipping the current mask to the MTC at run rollover. The rest of the cases are handled properly since SR settings are loading in the GUI when hitting Start or Resync.
One solution is use the run type word that is stored directly in the SR settings when rolling over. In this way the user is free to mess with the run word.

jcaravaca:
This could be a solution: reload the run type from the last at rollover. It works fine as far as I could test locally.

jcaravaca:
Tested OK at the TS.

tlatorre-uchicago:
One button that should be greyed out is the Trigger button on the caen GUI.

A couple other ones that aren't obvious what we should do are:

- MTC relay masks
- GT crates
- Pedestal crate mask

These are not part of the standard run and although it's not possible to load them directly from the GUI, it is possible to change them and then they will get changed when the run rolls over.

I think this is OK, but just wanted to bring it up to make sure.

The run type word trick has a race condition. We could have it use the standard run settings to reload the run type at every run start but perhaps until this issue #450 is worked out it's probably safest to leave it locked. The original motivation was to be able to edit the standard runs more easily, but I've been using the teststand to do that lately so it's not as big of a priority.

Last note is that the minor change in the CAEN nib file seems to have added a *huge* amount of lines to the file. Do you happen to know why that might be? Is it just a newer version of XCode you're using?

jcaravaca:
I think the important thing is not allow to change those in the middle of a run, so I think that, although probably not ideal, it should be fine...

Re. runTypeWord, since we rarely should change the run type word for an already defined standard run, it's fine if I leave it locked.

Not sure what's the deal with the .nib file. I'm in XCode 6.4. Can give a try to a different version at the teststand if there is a concern with this.

It'd be good to get this merged this week, so let me know if you see any other potential problem.

Source/Experiments/SNOP/SNOPController.m
```diff
@@ -1833,46 +1833,46 @@ - (void) runsLockChanged:(NSNotification*)aNotification
...
+    [debugDBClearButton setEnabled:!locked];
     [nhitMonitorAutoPulserRate setEnabled:!locked];
     [nhitMonitorAutoNumPulses setEnabled:!locked];
     [nhitMonitorAutoMaxNhit setEnabled:!locked];
     [nhitMonitorRunTypeWordMatrix setEnabled:!locked];
     [nhitMonitorCrateMaskMatrix setEnabled:!locked];
     [nhitMonitorTimeInterval setEnabled:!locked];
 
+    for(int irow=0;irow<21;irow++){
+        [[runTypeWordMatrix cellAtRow:irow column:0] setEnabled:notRunningOrInMaintenance];
```
tlatorre-uchicago: Should this be !lockedOrNotRunningMaintenance?

tlatorre-uchicago:
merged.

PR #441, opened 2017-08-16
check caen settings before starting standard run

tlatorre-uchicago:
This commit adds a check when the user starts a run to see if the CAEN settings for the new run are different than the current settings. If so, the run is resynced instead of doing a continuous run start.

### DIFF
--- a/Source/Experiments/SNOP/SNOPController.m
+++ b/Source/Experiments/SNOP/SNOPController.m
@@ -723,7 +723,6 @@ snopGreenColor;
 - (IBAction) startRunAction:(id)sender
 {
     /* Action when the user clicks on the Start or Restart Button. */
-    SNOCaenModel* caen;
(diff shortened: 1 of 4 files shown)


tlatorre-uchicago:
Tested on the teststand on August 16, 2017.

I verified that when starting from a cold start, the check is not performed. I also verified that switching from Maintenance -> Physics doesn't cause a resync, but that switching from Physics -> Smellie and from Smellie -> Physics does cause a resync.

The only potential issue with this solution is that I only check the model values and not the *true* hardware state. But, given that the hardware state is synced on every handover and it's very unlikely for a shifter to change the caen settings in the middle of a run this should be ok.

Ideally in the future we'd have a way to sync the caen state before the run.

@jcaravaca can you take a quick look at this and let me know what you think?

tlatorre-uchicago:
merged.

PR #444, opened 2017-08-17
load tubii settings at the start of continuous run starts

tlatorre-uchicago:
Currently TUBii settings are only loaded from hard run starts or after a resync. This commit updates it so that tubii settings are loaded for all run starts.


tlatorre-uchicago:
merged.

Issue #439, opened 2017-08-15
tubii ping doesn't acknowledge that the ping was received


tlatorre-uchicago:
In the TUBii GUI there is a "Ping" button, but when you press it it just says "Sending ping to TUBii" in the status window, but doesn't tell you whether it successfully connected or not.

icoulter:
I thought it wrote on the DAQ log. Also, I'm pretty sure it issues an error if there isn't a TUBii connected to that address.

tlatorre-uchicago:
It would be nice to just print "ping successful" or something. "Ping sent" sounds like it sent it, but never received a response.

icoulter:
Fair enough. I'll add it to the list and slip it in with the next firmware update.

tlatorre-uchicago:
I mean in the Orca dialog window. For example, on the teststand, I was trying to see if I successfully connected to the fake tubii server and this is what the status window shows when you click ping:

![screen shot 2017-08-16 at 5 06 46 pm](https://user-images.githubusercontent.com/5834059/29385339-50dbb012-82a5-11e7-9b4d-e7a49ceaebf2.png)

I think just after "Sending ping to TUBii" it should print if it successfully got a response back.

icoulter:
Yes. I'm agreeing with you. But the problem is that the firmware/orca aren't coordinating any response so it'll take a firmware update as well. I'll change it next time I need to do something else as its a pretty minor issue.

tlatorre-uchicago:
Fixed in #445.

PR #445, opened 2017-08-28
Add a response to ping if it manages to transmit successfully.

icoulter:


icoulter:
Should fix #439 
If the ping command doesn't get caught in the try statement, if prints a "success" comment to the DAQ log.

tlatorre-uchicago:
merged.

Issue #443, opened 2017-08-17
tellie does not stop firing when run changes


tlatorre-uchicago:
Today we were running TELLIE and forgot to stop firing the fibre when we switched into a physics run. The fibre continued to fire during the physics run.

The TELLIE code needs to post a run stop change wait notification before the run stops and stop firing before the run changes.

EdLeming:
Oh, I didn't see this. Thanks for tagging me. @tlatorre-uchicago, could you confirm the sequence of button clicks on this occation? 

The stopRunAction method of SNOPController.m does have a check for this already. That said, now I think about it there is also the situation where one could just change the standard run selection and hit the start / restart button. In that case the check I put in place will not be called.

tlatorre-uchicago:
I don't remember for sure, but we probably did click start run and not stop run.

ecaden:
I noticed this happening again when trying to recover from a breakdown/LV issue on crate 11 this morning.

EdLeming:
Yeah, If a user restarts / resyncs into a maintenance run with out explicitly hitting the stop *ellie button this will happen atm. I'll put up a PR later today.

tlatorre-uchicago:
Fixed in #446.

PR #446, opened 2017-08-31
Prevent ELLIE light in the detector when user enters another run without stopping a flash sequence

EdLeming:
Addresses issue #443. Copies the functionality already associated with the StopRun button into the Start / restart button.

Will test on the teststand today (31/08/17)


EdLeming:
Sucessfully tested on the teststand. I've copied below the ORCA log output observed when restarting into a maintenance when a SMELLIE sequence was running:

17-08-31 13:02:36 Starting Run 15379.9 (sub-run)
17-08-31 13:02:36 Sending SetSmellieDelay 200 to TUBii
17-08-31 13:02:37 Run 15379.9 preparing for a new sub-run
17-08-31 13:02:37 Starting Run 15379.10 (sub-run)
17-08-31 13:02:37 Sending SetSmellieDelay 200 to TUBii
17-08-31 13:02:38 CAEN settings are different in this standard run. Resyncing...
17-08-31 13:02:39 Standard run MAINTENANCE (DEFAULT) settings loaded. 
17-08-31 13:02:39 Prepare for new subrun ignored because run is stopping.
17-08-31 13:02:39 Start for new subrun ignored because run is stopping.
17-08-31 13:02:39 [SMELLIE]: thread has been cancelled, killing sequence.
17-08-31 13:02:39 [SMELLIE]: Sent to err statement. Stopping fire sequence.
17-08-31 13:02:39 [SMELLIE]: Stopped sending keep-alive to interlock server
17-08-31 13:02:39 [SMELLIE]: Smellie laser interlock server disarmed
17-08-31 13:02:39 Sending SetSmelliePulser 1 0.1 0 to TUBii
17-08-31 13:02:39 [SMELLIE]: Run sequence stopped.
17-08-31 13:02:39 DataTaking Thread Exited
17-08-31 13:02:39 Transfer queue cleared successfully
17-08-31 13:02:39 Waiting for transfer thread to exit
17-08-31 13:02:39 Transfer thread exited
17-08-31 13:02:39 Closing dataFile: ~/2017/08/Data/openFiles/Run15379
17-08-31 13:02:39 Move ~/2017/08/Data/openFiles/Run15379 --> ~/2017/08/Data/Run15379  <OK>
17-08-31 13:02:39 Final end of run cleanup
17-08-31 13:02:39 Run 15379 stopped.

tlatorre-uchicago:
The way to catch this in all cases is to look for the notification that the run is about to stop. Here is an example of how I did it for the nhit monitor: https://github.com/snoplus/orca/blob/3a5b00619bdb345d9b3d52915cd1b8333144acc3/Source/Experiments/SNOP/NHitMonitor.m#L202.

Essentially you need to catch the `ORAboutToQuit` and `ORAboutToStop` notifications, post a notification telling the run control to wait, and then wait until the threads are finished and post another notification telling the run control to continue.

icoulter:
A question on this:
I had thought that the standard runs would set the ellie pulsers to zero outside their specific runs. Maybe we haven't added that yet?

Source/Objects/Custom Hardware/SNO+/ELLIE/ELLIEController.m
```diff
@@ -123,6 +129,21 @@ -(void)fetchConfigurationFile:(NSNotification *)aNote{
     [model fetchCurrentSmellieConfig];
 }
 
+-(void)checkAndTidyELLIEThreads:(NSNotification *)aNote
```
tlatorre-uchicago: These should go in the respective models. It's possible to have more than one controller per model if you duplicate windows.

EdLeming:
@tlatorre-uchicago. Ah, that's cool. I've updated so it picks up that note. 

Seeing as this process is no longer associated with specific button pushed, I've also moved the observer / functionality to ELLIEModel.

@icoulter, that is indeed the case already. Unfortunately, due to the extremely large TTS on the light production from the superK laser SMELLIE always runs in master so this could not stop it.

icoulter:
Ah, makes sense. So this is only an issue for SMELLIE SK.

EdLeming:
@icoulter: for consistency we now just run all SMELLIE lasers in Slave, so it would have been a problem all around. 

@tlatorre-uchicago: Looks like your latest comment and my commit addressing it crossed in cyber space. I realised once I'd push it up they should have gone in the model instead.

EdLeming:
@tlatorre-uchicago. This PR, combined with changes I've made to the ELLIE server will now allow us to cancel SMELLIE blocking functions with a resolution of 5s. 

With this new functionality it is reasonable to extend the timeout of the smellie server client to some long time (currrently set to 20 mins). We can now run flash sequences upto 20 mins long, which we can cancel at any time. These changes addess issue #448

EdLeming:
@tlatorre-uchicago. This was tested UG (along with the server changes) during LV testing today. 

The new schema worked, however due to the long execution times of a number of the server hardware functions it still takes ~2 mins to fully stop a run. Running on the teststand with dummy servers the stop time is ~5s (See ORCA log output below). I'll speak to Jeff about doing what we can to speed it up when he's back from paternity leave.  

17-09-13 15:51:17 Run 15455 started(cold start).
17-09-13 15:51:17 DataTaking Thread Started
17-09-13 15:51:22 [SMELLIE]:Setting up a SMELLIE Run
17-09-13 15:51:22 [SMELLIE]: Running in MASTER mode
17-09-13 15:51:22 [SMELLIE]: Fire sequence requested for laser: PQ440
17-09-13 15:51:22 [SMELLIE]: Fire sequence requested for fibre: FS093
17-09-13 15:51:32 Run 15455 preparing for a new sub-run
17-09-13 15:51:32 Starting Run 15455.1 (sub-run)
17-09-13 15:51:42 Run 15455.1 preparing for a new sub-run
17-09-13 15:51:42 Starting Run 15455.2 (sub-run)
17-09-13 15:51:43 #############################################
17-09-13 15:51:43 [SMELLIE]
17-09-13 15:51:43 		Run stop button recongnised.
17-09-13 15:51:43 		EXTA triggers will stop within the next 5s.
17-09-13 15:51:43 		Putting the lasers to safe state may take
17-09-13 15:51:43 		upto 2 minutes.
17-09-13 15:51:43 #############################################
17-09-13 15:51:43 [SMELLIE]: Waiting for SMELLIE server to release blocking trigger function...
17-09-13 15:51:43 CAEN settings are different in this standard run. Resyncing...
17-09-13 15:51:43 Standard run MAINTENANCE (DEFAULT) settings loaded. 
17-09-13 15:51:43 [SMELLIE]: Waiting for SMELLIE server to release blocking trigger function...
17-09-13 15:51:47 [SMELLIE]: thread has been cancelled, killing sequence.
17-09-13 15:51:47 [SMELLIE]: Sent to err statement. Stopping fire sequence.
17-09-13 15:51:47 [SMELLIE]: Waiting for lasers to deactivate....
17-09-13 15:51:47 [SMELLIE]: Waiting for lasers to deactivate....
17-09-13 15:51:47 [SMELLIE]: Smellie laser interlock server disarmed
17-09-13 15:51:47 Sending SetSmelliePulser 1 0.1 0 to TUBii
17-09-13 15:51:47 [SMELLIE]: Stopped sending keep-alive to interlock server
17-09-13 15:51:47 [SMELLIE]: Smellie laser interlock server disarmed
17-09-13 15:51:47 Sending SetSmelliePulser 1 0.1 0 to TUBii
17-09-13 15:51:47 [SMELLIE]: Run sequence stopped.
17-09-13 15:51:47 [SMELLIE]: Run sequence stopped.
17-09-13 15:51:47 DataTaking Thread Exited
17-09-13 15:51:47 Transfer queue cleared successfully
17-09-13 15:51:47 Waiting for transfer thread to exit
17-09-13 15:51:47 Transfer thread exited
17-09-13 15:51:47 Closing dataFile: ~/2017/09/Data/openFiles/Run15455
17-09-13 15:51:47 Move ~/2017/09/Data/openFiles/Run15455 --> ~/2017/09/Data/Run15455  <OK>

Source/Experiments/SNOP/SNOPController.m
```diff
@@ -1547,11 +1528,23 @@ - (IBAction) startSmellieRunAction:(id)sender
...
 
 - (IBAction) stopSmellieRunAction:(id)sender
 {
+    if(![sender isKindOfClass:[NSNotification class]]){
+        NSLog(@"#############################################\n");
+        NSLog(@"[SMELLIE]\n");
+        NSLog(@"\t\tRun stop button recongnised.\n");
+        NSLog(@"\t\tEXTA triggers will stop within the next 5s.\n");
+        NSLog(@"\t\tPutting the lasers to safe state may take\n");
+        NSLog(@"\t\tupto 2 minutes.\n");
```
tlatorre-uchicago: typo upto -> up to

Source/Experiments/SNOP/SNOPController.m
```diff
@@ -1564,29 +1557,28 @@ - (IBAction) stopSmellieRunAction:(id)sender
...
 }
 
 - (IBAction) emergencySmellieStopAction:(id)sender
 {
+    NSLog(@"#############################################\n");
+    NSLog(@"[SMELLIE]\n");
+    NSLog(@"\t\tEmergency run stop button recongnised.\n");
+    NSLog(@"\t\tEXTA triggers will stop within the next [5s].\n");
+    NSLog(@"\t\tPutting the lasers to safe state may take\n");
+    NSLog(@"\t\tupto [2 minutes].\n");
```
tlatorre-uchicago: typo upto -> up to. What is the meaning of the times in square brackets?

Source/Objects/Custom Hardware/SNO+/ELLIE/ELLIEModel.m
```diff
@@ -2110,6 +2202,24 @@ -(void)smellieTidyUp
...
+        goto err;
+    }
+    SNOPModel* snopModel = [snopModels objectAtIndex:0];
+    
+    ////////////
+    // Roll over into maintenance run
+    NSLog(@"[SMELLIE]: Moving into a Maintenance run.\n");
+    if([[snopModel lastStandardRunType] isEqualToString:@"SMELLIE"]){
+        dispatch_sync(dispatch_get_main_queue(), ^{
+            [snopModel startStandardRun:@"MAINTENANCE" withVersion:@"DEFAULT"];
```
tlatorre-uchicago: Will this be called regardless of how the run is stopped? I'm wondering about the case that you start a new SMELLIE run while you are already in a SMELLIE run, i.e. you are in a SMELLIE run and decide it's not good for whatever reason or you want to start a new one, so you start a new smellie run. If I understand the logic correctly the old run will be stopped and then this will be called at the end starting a maintenance run.

I think what we want to happen is this when you are in a SMELLIE run:

1. If the run is explicitly stopped by the operator, then the run should just stop
2. If the operator starts a new run or resyncs, etc. then the run control should just start a new run or resync or whatever (with the caveat that the SMELLIE system will shut itself down before the new run is started)
3. If the SMELLIE run successfully finishes or encounters an error then it should move into a maintenance run.

I think the first two would be handled without any explicit action on the ELLIE model side. I think you could accomplish the third by starting a new standard run like you do here as long as it wasn't cancelled due to a run about to stop notice.

Source/Experiments/SNOP/SNOPController.m
```diff
@@ -1547,11 +1528,23 @@ - (IBAction) startSmellieRunAction:(id)sender
...
     //////////////////////
     // Start smellie thread
+    [theELLIEModel startInterlockThread];
     [theELLIEModel startSmellieRunThread:smellieRunFile];
+
 }
 
 - (IBAction) stopSmellieRunAction:(id)sender
 {
+    if(![sender isKindOfClass:[NSNotification class]]){
```
tlatorre-uchicago: Can you explain what this check does in a comment?

Source/Objects/Custom Hardware/SNO+/ELLIE/ELLIEModel.m
```diff
@@ -154,13 +157,17 @@ -(id) initWithCoder:(NSCoder *)decoder
 
         XmlrpcClient* tellieCli = [[XmlrpcClient alloc] initWithHostName:[self tellieHost] withPort:[self telliePort]];
         XmlrpcClient* smellieCli = [[XmlrpcClient alloc] initWithHostName:[self smellieHost] withPort:[self smelliePort]];
+        XmlrpcClient* smellieFlaggingCli = [[XmlrpcClient alloc] initWithHostName:[self smellieHost] withPort:[self smelliePort]];
```
tlatorre-uchicago: Needs to be released in `dealloc`.

EdLeming:
@tlatorre-uchicago, I think all comments are now addressed. The logic to cover your final point is a little more complicated than I would have liked (it requires two additional flags), but I've attempted to describe in comments.

Source/Experiments/SNOP/SNOPController.m
```diff
@@ -1400,6 +1412,7 @@ -(void) fetchSmellieRunFilesFinish:(NSNotification *)aNote
     [smellieLoadRunFile setEnabled:YES];
     
     [self setSmellieRunFileList:runFileDict];
+    NSLog(@"[SMELLIE] %i run plan documens sucessfully loaded\n", [runFileDict count]);
```
tlatorre-uchicago: typo documens -> documents

Source/Objects/Custom Hardware/SNO+/ELLIE/ELLIEModel.m
```diff
@@ -2110,7 +2212,38 @@ -(void)smellieTidyUp
...
+        NSArray*  snopModels = [[(ORAppDelegate*)[NSApp delegate] document] collectObjectsOfClass:NSClassFromString(@"SNOPModel")];
+        if(![snopModels count]){
+            NSLogColor([NSColor redColor], @"[TELLIE]: Couldn't find SNOPModel\n");
+            goto err;
+        }
+        SNOPModel* snopModel = [snopModels objectAtIndex:0];
+        if([[snopModel lastStandardRunType] isEqualToString:@"SMELLIE"]){
+            [snopModel startStandardRun:@"MAINTENANCE" withVersion:@"DEFAULT"];
+        }
+    } else { return; }
```
tlatorre-uchicago: Do you mean to have this else statement here? It looks like the code below will never be executed if maintenanceRollOver is set to false, and if it's true then you start a new standard run and then possibly start a new subrun?

Source/Objects/Custom Hardware/SNO+/ELLIE/ELLIEModel.m
```diff
@@ -2110,7 +2212,38 @@ -(void)smellieTidyUp
...
+    /////////////////////////////////////////
+    // If the sequence finished without external influence move into a maintenance run.
+    if([self maintenanceRollOver]){
+        NSArray*  snopModels = [[(ORAppDelegate*)[NSApp delegate] document] collectObjectsOfClass:NSClassFromString(@"SNOPModel")];
+        if(![snopModels count]){
+            NSLogColor([NSColor redColor], @"[TELLIE]: Couldn't find SNOPModel\n");
+            goto err;
+        }
+        SNOPModel* snopModel = [snopModels objectAtIndex:0];
+        if([[snopModel lastStandardRunType] isEqualToString:@"SMELLIE"]){
```
tlatorre-uchicago: Will this function ever get called outside of a smellie run? If not you can probably remove this check, otherwise I think it would be better to rely on the run type word which you can access via `[gOrcaGlobals runType]` to check the SMELLIE bit since someone may in the future create a SMELLIE standard run with a different name.

tlatorre-uchicago:
To make my idea above more concrete, here is what I meant:

```objective-c
- (void) startSmellieRun
{
    ...

    /* Stop the SMELLIE laser. */
    [self stopSmellie];

    if ([[NSThread currentThread] isCancelled]) {
        /* Run was stopped, started, or resynced, so we don't need to roll over into a new run. */
    } else {
        dispatch_async(dispatch_get_main_queue(), ^{
            [aSNOPModel startStandardRun:@"MAINTENANCE"];
        });
    }

err:
    [self stopSmellie];
}

- (void) stopSmellie
{
    /* This is a blocking function which ensures that SMELLIE is not running. */
    [self deactivateSmellie];
    [self killKeepAlive:nil];
    [self stopSmelliePulser];
}

- (void) runAboutToStop: (NSNotification *) aNote
{
    /* Run was stopped externally, i.e. someone hit stop, start, or
     * resync. In this case, all that is necessary is to stop everything and
     * the run control will handle stopping or moving into a new run. */
    if ([smellieThread isRunning]) {
        // Here, you post the notification telling the run control to wait and cancel the thread. No need to handle new runs.
        [smellieThread cancel];
        [NSThread detachNewThreadSelector:@selector(waitForSmellieRunToFinish) toTarget:self withObject:nil];
    }
}

-(void) waitForSmellieRunToFinish
{
    @autoreleasepool {
        while ([[self smellieThread] isExecuting]) {
        }

        /* Go ahead and end the run. */
        [[NSNotificationCenter defaultCenter] postNotificationOnMainThreadWithName:ORReleaseRunStateChangeWait object: self];
    }
}
```

EdLeming:
hi @tlatorre-uchicago. So I think I've addressed your points.

 You're right about tidying up the sequence, it was overly complicated. I've updated both the SMELLIE and TELLIE "start*ELLIERun" functions to include the shut down commands. The "stop*ELLIERun" functions now just check if the threads are running. If they are, the functions cancels them and launches a thread that waits until they have finished and then handles the run transition.

I've also taken all the GUI changes out. I'll add them back in once Andy (and Javi) have merged.

tlatorre-uchicago:
Looks good to me. I think just a couple of things and I'll merge ASAP:

- few typos that I commented on earlier
- one comment that would be nice that I commented earlier
- delete extra whitespace

EdLeming:
@tlatorre-uchicago, sorry about the missing typo corrections etc. I did put them in a commit a while back, but it looks like they got removed when I reverted the commit with the SNOP.xib changes. 

I've also noticed a few flags associated with handling run transitioning were removed at the same time. These have been added back (heavily commented) and the schema successfully re-tested.

In making these changes I have dug up one bug that I cannot for the life of me get to the bottom of. Some advice would be appreciated. When I hit the smellieEmergencyStop button the button goes blue and beach balls. Putting break points and stepping though the associated action function line by line suggests everything executes as expected. The ELLIEModel function that gets called as a result of a notification (posted to the main thread) also runs fine. I'm not sure how to go about further de-bugging.

EdLeming:
@tlatorre-uchicago. It appears to be some issue with catching posted notifications. The only advice I can find for this online is to ensure there was no object which was registered as an observer and deleted without the observers being released. In this case the only observer was in the ELLIEModel which already removes these observers on dealloc.

My solution is to call the killKeepAlive method directly. I don't expect any of the server calls in this function to block for any macroscopic amount of time. They should both simply set flags on the respective servers, then return. As a result, calling these functions directly comes with a very small probablity of blocking GUI interactions.

tlatorre-uchicago:
merged.

PR #452, opened 2017-09-27
Database lock to prevent multiple copies of Orca controlling SNO+

mastbaum:
In order to prevent multiple copies of Orca from controlling SNO+ at once, this adds an exclusive locking system based on advisory locks in PostgreSQL.

When initializing the `SNOPModel`, new logic will connect to a user-defined DB (configured in a new UI in the SNOPModel settings pane) and obtain a lock. If it succeeds, Orca starts as usual and the lock is held as long as Orca is open. If it fails, a dialog box will pop up alerting the user and giving the following options:

* Quit Orca immediately
* Ignore locks and open Orca anyway
* Try again to obtain the lock

In the background, Orca will try every five seconds to obtain the lock. If it succeeds, the dialog will close and Orca will proceed to load.

Currently the locking is aggressive, in that it will block `init`, `initWithCoder`, or `awakeAfterDocumentLoaded`. In principle only the latter is needed, but there is a lot of logic in `initWithCoder` and blocking the init methods assures that Orca doesn't talk to any DB without a lock first.

If there is no database address specified in the Orca file being loaded, locking is disabled.


Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -120,6 +128,9 @@ - (id) init
     start = COLD_START;
     resync = NO;
 
+    /* Check that not other Orca is controlling the detector, using a database lock. This will either block or quit. */
```
tlatorre-uchicago: typo "Check that not other Orca is controlling the detector" -> "Check that other Orca is not controlling the detector"

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -146,6 +157,107 @@ - (id) init
...
+                               dataBase:_lockDBName];
+    }
+
+    /* Try to get an advisory lock, returns t if we get it or already own it, else f. */
+    if ([[self dbLockConnection] isConnected]) {
+        ORPQResult* result = [dbLockConnection queryString:[NSString stringWithFormat:@"select pg_try_advisory_lock(%i);", _lockDBLockID]];
+        gotLock = [[result fetchRowAsType:MCPTypeDictionary row:0] valueForKey:@"pg_try_advisory_lock"];
+        NSLog(@"Obtaining lock for %@ on %@:%u/%@ ID %u. Result: %@\n",
+              _lockDBUserName, _lockDBIPAddress, _lockDBPort, _lockDBName, _lockDBLockID, gotLock);
+    }
```
tlatorre-uchicago: Should probably warn here that it was unable to connect to the database, otherwise the user will have no way of knowing that they aren't connecting to the database.

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -146,6 +157,107 @@ - (id) init
...
+        ORAppDelegate* delegate = [NSApp delegate];
+        [delegate terminate:self];
+    }
+    else if (modalAction == NSAlertSecondButtonReturn) {
+        [alert release];
+        [self checkDatabaseLock];
+    }
+    else if (modalAction == NSAlertThirdButtonReturn) {
+        ignoreDBLock = YES;
+    }
```
tlatorre-uchicago: Need to add an else block if the modal was aborted from timerCheckDatabaseLock?

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -146,6 +157,107 @@ - (id) init
...
+    if (modalAction == NSAlertFirstButtonReturn) {
+        ORAppDelegate* delegate = [NSApp delegate];
+        [delegate terminate:self];
+    }
+    else if (modalAction == NSAlertSecondButtonReturn) {
+        [alert release];
+        [self checkDatabaseLock];
+    }
+    else if (modalAction == NSAlertThirdButtonReturn) {
+        ignoreDBLock = YES;
```
tlatorre-uchicago: need to release alert here

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -146,6 +157,107 @@ - (id) init
...
+
+    [[NSRunLoop currentRunLoop] addTimer:dbModalTimer forMode:NSModalPanelRunLoopMode];
+
+    /* Launch the modal. */
+    [[NSRunningApplication currentApplication] activateWithOptions:NSApplicationActivateIgnoringOtherApps];
+    int modalAction = [alert runModal];
+
+    /* Respond to a press of the modal buttons. */
+    if (modalAction == NSAlertFirstButtonReturn) {
+        ORAppDelegate* delegate = [NSApp delegate];
```
tlatorre-uchicago: need to release alert here

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -280,6 +392,17 @@ - (id) initWithCoder:(NSCoder*)decoder
...
+    // Database lock
+    self.lockDBUserName = [decoder decodeObjectForKey:@"ORSNOPModelLockDBUserName"];
+    self.lockDBPassword = [decoder decodeObjectForKey:@"ORSNOPModelLockDBPassword"];
+    self.lockDBName = [decoder decodeObjectForKey:@"ORSNOPModelLockDBName"];
+    self.lockDBIPAddress = [decoder decodeObjectForKey:@"ORSNOPModelLockDBIPAddress"];
+    self.lockDBPort = [decoder decodeInt32ForKey:@"ORSNOPModelLockDBPort"];
+    self.lockDBLockID = [decoder decodeInt32ForKey:@"ORSNOPModelLockDBLockID"];
+
+    /* Check that not other Orca is controlling the detector, using a database lock. This will either block or quit. */
+    [self checkDatabaseLock];
```
tlatorre-uchicago: I think this should only go in the awakeAfterDocumentLoaded section

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -120,6 +128,9 @@ - (id) init
     start = COLD_START;
     resync = NO;
 
+    /* Check that not other Orca is controlling the detector, using a database lock. This will either block or quit. */
+    [self checkDatabaseLock];
```
tlatorre-uchicago: I think this should only go in the awakeAfterDocumentLoaded function

mastbaum: I tried that, but this means that models get initialized, and I was worried about side effects in  those initialization methods. Database connections are made, for example querying the HV settings from the `ORXL3Model`, and so pre-empting model initialization seemed to be the only way guarantee that the lock prevents Orca from doing anything. But if everyone else is contractually obligated not to control hardware or write to the DB during `init`/`initWithCoder`, deferring this to `awakeAfterDocumentLoaded` would work too (and avoid loading time overhead during hand-offs). Or is there something else that makes it safe?

tlatorre-uchicago: I don't know of any guaranteed ordering in the initialization of models, i.e. I don't think that you are guaranteed to block these models from accessing the database by putting this check in initWithCoder. I think putting it in a single place has a couple of advantages:

1. The rest of Orca is guaranteed to be up and running so when the handover happens there is minimal downtime.

2. It's easier to understand the order of operations.

mastbaum:
Thanks for the feedback @tlatorre-uchicago! I've updated per review comments in 9d8a38d:

* Check the lock in `awakeAfterDocumentLoaded`, not initialization methods (`init` and `initWithCoder`)
* Red log warning if unable to obtain the DB lock
* Release the alert object when closed by modal actions

tlatorre-uchicago:
@mastbaum, I have created a new database on both the test database on minard and the detector state database on dbug. The database name is "orca" and it has the following table:

```sql
CREATE TABLE orca_sessions (
    key                 bigserial PRIMARY KEY,
    timestamp           timestamp with time zone default now(),
    end_timestamp       timestamp with time zone,
    ip_addr     inet DEFAULT inet_client_addr(),
    orca_version    text,
    hostname        text,
    osx_version     text
);
```

The operator user should be able to insert and update the columns.

Ideally Orca would insert a row when it successfully grabs the lock and then updates the end_timestamp when Orca is about to quit. Let me know if you want to add this functionality to this pull request or wait. It might be a good idea to have so that if we ever end up in a situation where someone has the lock we can quickly debug where it is.

Getting the hostname should be easy: https://stackoverflow.com/questions/15368613/how-do-i-get-hostname-of-ios-device.

You can get the version of OSX also pretty easily: https://stackoverflow.com/questions/3694940/get-osx-version-with-objective-c.

The orca version is currently not defined anywhere in the code, but I think it would be fine to add a #define in the SNOPModel.h file.

Source/Objects/Misc Objects/DatabaseSupport/PostgreSQL/ORPQConnection.m
```diff
@@ -139,8 +139,8 @@ - (BOOL) checkConnection
 {
 	BOOL result = NO;
 	@synchronized(self){
-		PGresult *res = PQexec(mConnection,"");
-		if (PQresultStatus(res) == PGRES_COMMAND_OK) {
+		PGresult *res = PQexec(mConnection,"select 1 limit 0");
+		if (PQresultStatus(res) == PGRES_TUPLES_OK) {
```
mastbaum: @boardhead I found while testing that `checkConnection` always returned `NO`, because for a null query `PQresultStatus` returned `PGRES_EMPTY_QUERY` rather than `PGRES_COMMAND_OK` (see e.g. [here](https://doxygen.postgresql.org/libpq-fe_8h.html#a16a9effd60c2048ab36ab3bf52bede2ca5bcebb782e39c46a1f6fe2e1c17634a7)). Since I wasn't positive that an empty query will always be executed on the server (it seems plausible that the client could just immediately return `PGRES_EMPTY_QUERY`), I replaced it with this tuple-returning no-op.

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -438,6 +632,21 @@ - (void) sleep
 
 - (void) awakeAfterDocumentLoaded
 {
+    /* Check that no other Orca is controlling the detector, using a database lock. This will either block or quit. */
+    [self checkDatabaseLock:true];
+
+    [self postSessionStart];
+
+    /* A timer to check the lock every 10 seconds, to make sure we still have it. */
+    //NSTimer* dbCheckTimer =
```
tlatorre-uchicago: delete?

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -438,6 +632,21 @@ - (void) sleep
...
+    [NSTimer scheduledTimerWithTimeInterval:10.0
+                                                    target:self
+                                                  selector:@selector(timerCheckDatabaseLock:)
+                                                  userInfo:nil
+                                                   repeats:YES];
+
+    //[[NSRunLoop mainRunLoop] addTimer:dbCheckTimer forMode:NSRunLoopCommonModes];
```
tlatorre-uchicago: delete?

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -146,6 +156,182 @@ - (id) init
...
+{
+    bool connected = [self initOrcaSessionDBConnection];
+
+    if (!connected) {
+        NSLogColor([NSColor redColor], @"Unable to post session end to DB: DB not connected.");
+        return;
+    }
+
+    NSString* query = [NSString stringWithFormat:@"update orca_sessions set end_timestamp = now() where key = %@", [self sessionKey]];
+    NSLog(@"%@", query);
```
tlatorre-uchicago: need a newline or maybe just delete this?

tlatorre-uchicago:
merged.

PR #453, opened 2017-10-04
update the mapping between crates and channels on the MTCA+ at run starts

tlatorre-uchicago:
This commit updates the MTC model to send a command to update the mapping between crates and channels on the MTCA+s at the start of every run.


tlatorre-uchicago:
merged.

PR #454, opened 2017-10-06
ECA modes

jcaravaca:
Supersedes #390. Changes are:
- Remove 'Start Single ECA' and campaign mode
- Select ECA from the SR dropdown menu to start an ECA. There are 2 different modes:
-- User settings: load the settings in the GUI
-- ECA standard: load hardcoded values
- The ECA mode is taken from the run type word (ECA, Supernova or Physics)
- The ECA type is taken from the run type word (PDST or TSLP)
- I didn't figure out how to schedule a new SR after the ECA so now the run doesn't stop when the ECA routine is done. It's up to the user to go to the next run.
- I left a place holder for a scheduler of a next SR. This doesn't work as is, since there is a race condition: runStopped and runPostStopped are not guaranteed to run when the run is fully stop, 'restartRun' is sometimes called while the run is stopping and ORCA doesn't handle this properly. Maybe there is a way to circumvent this, but I didn't figure that out and at this point it's out of the scope of this PR.

### DIFF
--- a/Source/Experiments/SNOP/ECARun.m
+++ b/Source/Experiments/SNOP/ECARun.m
@@ -106,6 +106,7 @@ NSString* ORECARunFinishedNotification = @"ORECARunFinishedNotification";
                                 exceptionWithName:@"ECASettings"
                                 reason:@"ECA settings not set"
                                 userInfo:nil];
+    
     @try{
         isFinished = NO;
         start_eca_run = NO;
@@ -118,23 +119,6 @@ NSString* ORECARunFinishedNotification = @"ORECARunFinishedNotification";
         }
         aSNOPModel = [objs objectAtIndex:0];
 
-        // Store previous runs to rollover at the end
-        // only if CONTINUOUS_START
-        NSLogColor([NSColor redColor], @"setECASettings: startmode %i.\n", [aSNOPModel startMode]);
-        if( [aSNOPModel startMode] == 1){
-            previousSR = [[aSNOPModel lastStandardRunType] copy];
-            previousSRVersion = [[aSNOPModel lastStandardRunVersion] copy];
-            [aSNOPModel setNextStandardRunType:previousSR];
-            [aSNOPModel setNextStandardRunVersion:previousSRVersion];
-            NSLogColor([NSColor redColor], @"setECASettings: %@ and %@ \n",previousSR,previousSRVersion);
-        }
-        prev_gtmask = [anMTCModel GTCrateMask];
-        prev_pedmask = [anMTCModel pedCrateMask];
-        prev_coarsedelay = [anMTCModel coarseDelay];
-        prev_finedelay = [anMTCModel fineDelay];
-        prev_pedwidth = [anMTCModel pedestalWidth];
-
-        /* Check run type word and set parameters accordingly */
         objs = [[(ORAppDelegate*)[NSApp delegate] document] collectObjectsOfClass:NSClassFromString(@"ORRunModel")];
         if (![objs count]) {
             NSLogColor([NSColor redColor], @"setECASettings: Run control not found.\n");
@@ -142,35 +126,6 @@ NSString* ORECARunFinishedNotification = @"ORECARunFinishedNotification";
         }
         ORRunModel *aRunModel = [objs objectAtIndex:0];
         unsigned long runTypeWord = [aRunModel runType];
-        //Set whether PDST or TSLP
-        if( (runTypeWord & kECAPedestalRun) && (runTypeWord & kECATSlopeRun) ){
-            NSLogColor([NSColor redColor], @"setECAsettings: Both PDST and TSLP bits checked in the run type word. Don't know what to do!.\n");
-            @throw ECAException;
-        }
-        else if( runTypeWord & (kECAPedestalRun) ){
-            [self setECA_type:@"PDST"];
-        }
-        else if( runTypeWord & (kECATSlopeRun) ){
-            [self setECA_type:@"TSLP"];
-        }
-        else{
-            NSLogColor([NSColor redColor], @"setECAsettings: not an ECA run.\n");
-            @throw ECAException;
-        }
-        //Set mode
-        if( runTypeWord & (kECARun) ){
-            [self setECA_mode:ECAMODE_DEDICATED];
-        }
-        else if( runTypeWord & (kSupernovaRun) ){
-            [self setECA_mode:ECAMODE_SUPERNOVA];
-        }
-        else if( runTypeWord & (kPhysicsRun) ){
-            [self setECA_mode:ECAMODE_PHYSICS];
-        }
-        else{
-            NSLogColor([NSColor redColor], @"setECAsettings: Not a valid run for ECA_mode.\n");
-            @throw ECAException;
-        }
         //Load hardcoded values if requested
         if([[aSNOPModel standardRunVersion] isEqualToString:@"ECA_PDST_STANDARD"]){
             //Pedestal run
@@ -187,12 +142,62 @@ NSString* ORECARunFinishedNotification = @"ORECARunFinishedNotification";
             [self setECA_nevents:ECA_TSLP_NEVENTS];
         }
         else if([[aSNOPModel standardRunVersion] isEqualToString:@"ECA_USER_SETTINGS"]){
-            NSLogColor([NSColor redColor], @"setECASettings: using user input.\n");
+            if ([ECA_type isEqualToString:@"PDST"]) {
+                runTypeWord = ~(~runTypeWord | kECATSlopeRun);
+                runTypeWord |= kECAPedestalRun;
+                [aRunModel setRunType:runTypeWord];
+            }
+            else if([ECA_type isEqualToString:@"TSLP"]) {
+                runTypeWord =~(~runTypeWord | kECAPedestalRun);
+                runTypeWord |= kECATSlopeRun;
+                [aRunModel setRunType:runTypeWord];
+            }
         }
         else{
             NSLogColor([NSColor redColor], @"setECASettings: Unknown standard run version.\n");
             @throw ECAException;
         }
+        
+        /* Check run type word and set parameters accordingly */
+        if( (runTypeWord & kECAPedestalRun) && (runTypeWord & kECATSlopeRun) ){
+            NSLogColor([NSColor redColor], @"setECAsettings: Both PDST and TSLP bits checked in the run type word. Don't know what to do!.\n");
+            @throw ECAException;
+        }
+        else if( runTypeWord & (kECAPedestalRun) ){
+            [self setECA_type:@"PDST"];
+        }
+        else if( runTypeWord & (kECATSlopeRun) ){
+            [self setECA_type:@"TSLP"];
+        }
+        else{
+            //Not an ECA run
+            @throw ECAException;
+        }
+        //Set mode
+        if( runTypeWord & (kECARun) ){
+            [self setECA_mode:ECAMODE_DEDICATED];
+        }
+        else if( runTypeWord & (kSupernovaRun) ){
+            [self setECA_mode:ECAMODE_SUPERNOVA];
+        }
+        else if( runTypeWord & (kPhysicsRun) ){
+            [self setECA_mode:ECAMODE_PHYSICS];
+        }
+        else{
+            NSLogColor([NSColor redColor], @"setECAsettings: Not a valid run for ECA_mode.\n");
+            @throw ECAException;
+        }
+
+        // If we are here, that means this is an ECA run. Do the thing:
+        // Start a maintenance run with previous settings after ECAs are done
+        [aSNOPModel setNextStandardRunType:@"MAINTENANCE"];
+        [aSNOPModel setNextStandardRunVersion:@"DEFAULT"];
+        prev_gtmask = [anMTCModel GTCrateMask];
+        prev_pedmask = [anMTCModel pedCrateMask];
+        prev_coarsedelay = [anMTCModel coarseDelay];
+        prev_finedelay = [anMTCModel fineDelay];
+        prev_pedwidth = [anMTCModel pedestalWidth];
+        
         /* Set pulser rate:
          * If we starting an ECA run we have to set the pulser
          * rate enter by the operator, not the standard run one */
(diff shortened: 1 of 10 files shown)


jcaravaca:
Tested at the TS.

Source/Experiments/SNOP/ECARun.m
```diff
@@ -7,23 +7,43 @@
...
+NSString* ORECAStatusChangedNotification = @"ORECAStatusChangedNotification";
 NSString* ORECARunChangedNotification = @"ORECARunChangedNotification";
 NSString* ORECARunStartedNotification = @"ORECARunStartedNotification";
 NSString* ORECARunFinishedNotification = @"ORECARunFinishedNotification";
 
-//Definitions
+/* This flag must be enabled for commissioning 
+ * and normal detector running. Disable it only 
+ * for testing. */
+//#define __COMMISSIONING__ 1
```
tlatorre-uchicago: This shouldn't be uncommented right? The name is very misleading, commissioning sounds like testing which is the opposite of what it does.

Source/Experiments/SNOP/ECARun.m
```diff
@@ -123,54 +89,175 @@ - (BOOL) isExecuting
...
+        /* Set pulser rate:
+         * If we starting an ECA run we have to set the pulser
+         * rate enter by the operator, not the standard run one */
+        objs = [[(ORAppDelegate*)[NSApp delegate] document] collectObjectsOfClass:NSClassFromString(@"ORMTCModel")];
+        if (![objs count]) {
+            NSLogColor([NSColor redColor], @"setECASettings: MTC not found.\n");
+            @throw ECAException;
+        }
+        anMTCModel = [objs objectAtIndex:0];
+        [anMTCModel setGtMask:( [anMTCModel gtMask] | MTC_EXT_8_MASK )];
```
tlatorre-uchicago: The GT mask should be set by the standard run type.

Source/Experiments/SNOP/ECARun.m
```diff
@@ -123,54 +89,175 @@ - (BOOL) isExecuting
...
+    [ECAThread start];
+    //[ECAThread autorelease];
+    [[NSNotificationCenter defaultCenter] postNotificationOnMainThreadWithName:ORECARunStartedNotification object:self];
+
 }
 
 - (void) doECAs
 {
+
+    /* This is the actual ECA algorithm, that runs must run a different thread */
```
tlatorre-uchicago: typo? does this mean to say that "this function runs on a separate thread"?

Source/Experiments/SNOP/ECARun.m
```diff
@@ -195,65 +282,75 @@ - (void) doECAs
...
-            if(([anMTCModel gtMask] & MTC_EXT_8_MASK) || !([anMTCModel gtMask] & MTC_PULSE_GT_MASK)){
-                NSLogColor([NSColor redColor], @" Enable PGT and disable EXT PED bit in the MTC trigger mask. Stopping ECAs... \n");
-                goto stop;
-            }
-
             if(![self doPedestals]) {
                 // User stopped manually
-                start_new_run = false;
+                [aSNOPModel setNextStandardRunType:nil];
+                [aSNOPModel setNextStandardRunVersion:nil];
```
tlatorre-uchicago: Since this isn't working in this PR, these should be removed.

Source/Experiments/SNOP/ECARun.m
```diff
@@ -195,65 +282,75 @@ - (void) doECAs
...
-            if(!([anMTCModel gtMask] & MTC_EXT_8_MASK)){
-                NSLogColor([NSColor redColor], @" Enable EXT PED bit in the MTC trigger mask. Stopping ECAs... \n");
-                goto stop;
-            }
-
             if(![self doTSlopes]) {
                 // User stopped manually
-                start_new_run = false;
+                [aSNOPModel setNextStandardRunType:nil];
+                [aSNOPModel setNextStandardRunVersion:nil];
```
tlatorre-uchicago: same here.

Source/Experiments/SNOP/SNOPController.m
```diff
@@ -780,6 +791,10 @@ - (IBAction) stopRunAction:(id)sender
         [theELLIEModel stopSmellieRun];
     }
 
+    //Make sure we don't start a new run
+    [model setNextStandardRunType:nil];
+    [model setNextStandardRunVersion:nil];
```
tlatorre-uchicago: same here.

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -117,7 +117,7 @@ - (id) init
 
     rolloverRun = NO;
     state = STOPPED;
-    start = COLD_START;
+    startMode = COLD_START;
```
tlatorre-uchicago: Why was this name changed?

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -271,13 +271,18 @@ - (NSString *) xl3Host
     return xl3Host;
 }
 
+- (int) startMode
+{
+    return startMode;
```
tlatorre-uchicago: This is just an internal state variable that shouldn't be accessed outside of the SNO+ model.

Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
```diff
@@ -1914,46 +1945,11 @@ - (void) loadTriggers
...
-    MultiSetCrateTriggersArgs *args = (MultiSetCrateTriggersArgs *) payload;
-
-    args->slotMask = [self getSlotsPresent];
-
-    for (slot = 0; slot < 16; slot++) {
-        args->tr100Masks[slot] = 0;
-        args->tr20Masks[slot] = 0;
+        unsigned int mask = [[aXL3Mask objectAtIndex:slot] unsignedIntValue];
+        args->tr100Masks[slot] &= mask;
+        args->tr20Masks[slot] &= mask;
```
tlatorre-uchicago: Can you leave the existing functions as is and add a new function to load the triggers with a given mask?

Source/Objects/Custom Hardware/SNO+/XL3/ORXL3Model.m
```diff
@@ -1970,8 +1966,8 @@ - (void) disableTriggers
 
     if (ntohl(results->errorMask)) {
         NSException *e = [NSException exceptionWithName:@"loadTriggersError"
-                          reason:@"failed to load channel triggers"
-                          userInfo:nil];
+                                                 reason:@"failed to load channel triggers"
+                                               userInfo:nil];
```
tlatorre-uchicago: Not sure why the formatting was changed here.

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -876,17 +897,23 @@ - (void) runStarted:(NSNotification*)aNote
...
-         * the beginning of the run to look for any trigger issues. */
+    if(run_type & (kECAPedestalRun | kECATSlopeRun)){
+        /* Launch the ECA routine if this is an ECA run, as in
+         * the PDST or TSLP bits are enabled */
+        [[self anECARun] launchECAThread:NULL];
+    }
+    else if (run_type & kPhysicsRun) {
+        /* If this is a physics run with no ECAs, we ping each slot in the 
+         * detector once at the beginning of the run to look for any trigger 
+         * issues. */
```
tlatorre-uchicago: We should double check with analysis, run selection, and DWG that these changes are clear, i.e. currently the physics run type word signifies that it's a good physics run and that no settings were changed during the run. With this change the semantics will be different, i.e. it will be necessary to check not just that the physics run bit is set but also that the ECA bits are not set.

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -576,23 +588,27 @@ - (void) runInitialization:(NSNotification*)aNote
...
         } else {
-            start = CONTINUOUS_START;
+            NSLogColor([NSColor redColor], @"runInitialization: CONTINUOUS_START.\n");
+            startMode = CONTINUOUS_START;
         }
         break;
     default:
-        start = COLD_START;
+        NSLogColor([NSColor redColor], @"runInitialization: COLD_START.\n");
+        startMode = COLD_START;
```
tlatorre-uchicago: leftover debugging?

Source/Experiments/SNOP/ECARun.m
```diff
@@ -123,54 +89,175 @@ - (BOOL) isExecuting
...
+
+        /* Enable Pedestals */
+        [anMTCModel setIsPedestalEnabledInCSR:TRUE];
+
+        /* Enable Pedestal and GT crate mask for all the crates */
+        [anMTCModel setGTCrateMask: (prev_gtmask | 0x7FFFF) ];
+        [anMTCModel setPedCrateMask: (prev_pedmask | 0x7FFFF) ];
+#ifdef __COMMISSIONING__
+        [anMTCModel loadGTCrateMaskToHardware];
+        [anMTCModel loadPedestalCrateMaskToHardware];
```
tlatorre-uchicago: Is there a reason you both set these in the GUI and load them to hardware in this function? It looks like it is only called from the SNO+ model before the run is initialized. In that case you should only have to set them in the GUI. Alternatively, you could just set them when the ECA starts during the run. I think the only thing that is necessary to set beforehand is the run type word.

Source/Experiments/SNOP/ECARun.m
```diff
@@ -123,54 +89,175 @@ - (BOOL) isExecuting
...
+    return isFinished;
 }
 
 - (BOOL) isFinishing
 {
     return isFinishing;
 }
 
-- (void) setPulserRate:(NSNotification*)aNote
+- (bool) setECASettings:(NSNotification*)aNote
```
tlatorre-uchicago: This function is defined as if it were to be called by a notification, but it's only called from the SNO+ model with NULL. Should it just be `- (bool) setECASettings`?

Source/Experiments/SNOP/ECARun.m
```diff
@@ -123,54 +89,175 @@ - (BOOL) isExecuting
...
+    if([ECAThread isExecuting]) {
+        NSLogColor([NSColor redColor], @"launchECAThread: ECA already running... \n");
+        return;
     }
 
+    //Launch ECA thread
+    [ECAThread release];
+    ECAThread = [[NSThread alloc] initWithTarget:self selector:@selector(doECAs) object:nil];
+    [ECAThread start];
+    //[ECAThread autorelease];
```
tlatorre-uchicago: delete?

Source/Experiments/SNOP/ECARun.m
jcaravaca: This is called by runAboutToStart, so the HW has already been initialized and changes in GUI won't have any effect. Is there any reason we don't want to prepare the HW for ECAs at this stage?

Source/Experiments/SNOP/SNOPModel.m
jcaravaca: XCode didn't like that 'start' was overloaded with this variable name and the start method. It was messing up with the 'nextStandardRun' logic, which will actually disappear. I'd suggest to keep this name change for the future, just in case.

Source/Experiments/SNOP/SNOPModel.m
jcaravaca: I was using this for the rollover logic of the ECA code, which has disappeared. So I'll take this out.

Source/Experiments/SNOP/SNOPModel.m
jcaravaca: I agree we should have a discussion with DQ/Analysis. I would suggest that DQ looks at these bits to understand if it was a pure physics run and reject them in case they determine they have an impact in physics. They are already doing so with the DQ bits, so this should be the same. Eventually we'll have also the 'Embedded pedestals bit', which will potentially by enabled in physics running as well.

jcaravaca:
Merged with master sorting out the GUI changes. Let me know if there is anything else to be addressed, otherwise I'll test it once more at the TS.

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -1047,7 +1047,12 @@ - (void) runAboutToStart:(NSNotification*)aNote
         goto err;
     }
 
-    switch (start) {
+    /* Load the ECA settings if it's an ECA run
+     * This will set MTC settings for correct ECA running, so 
+     * we want this to be done before loading settings in HW. */
+    [[self anECARun] setECASettings];
```
tlatorre-uchicago: This should only be called when an ECA run is starting right?

Source/Experiments/SNOP/ECARun.m
```diff
@@ -123,54 +89,171 @@ - (BOOL) isExecuting
...
+        [anMTCModel setPedCrateMask: (prev_pedmask | 0x7FFFF) ];
+#ifndef __TESTING__
+        [anMTCModel loadGTCrateMaskToHardware];
+        [anMTCModel loadPedestalCrateMaskToHardware];
+#endif
+
+        start_eca_run = YES;
+        return true;
+    }
+    @catch(...){
```
tlatorre-uchicago: I'd recommend a goto block here instead of a global catch all. The reason is that when reading the code it appears as though this function raises exceptions (it defines an exception at the top and then in many places throws an exception), so I thought at first that any calling code would need a try except. But since all these exceptions are meant to do as far as I can tell is to bring you down to the end of the function, I think it'd be easier to read if it were just a simple goto statement.

Source/Experiments/SNOP/SNOPModel.m
tlatorre-uchicago: I just realized that eventually if we want these runs to be real "physics" golden list runs we'll have to figure out a way to have ping crates run before the ECAs because they are currently a part of run selection. Not critical for this pull request, but something to think about.

Source/Experiments/SNOP/SNOPModel.m
tlatorre-uchicago: Also, based on the comment above, I think this is in the wrong place. First of all, the function shouldn't actually set any hardware settings if it's just setting things up before hardware settings are loaded.

I think the best place to call this is in the loadStandardRun method. As it is now, it will get called outside of the normal sequence of setting hardware values which could get confusing.

tlatorre-uchicago:
merged.

PR #459, opened 2017-10-31
SNO+: only load ECA settings if it's an ECA run

tlatorre-uchicago:
This pull request adds a check so that the ECA settings are only loaded if it's an ECA run.


tlatorre-uchicago:
merged.

PR #468, opened 2017-11-21
fix bug in setECASettings

tlatorre-uchicago:
This commit fixes a bug in the new ECA code which was using an uninitialized variable when accessing the MTC object at the start of the run.

In addition, I took out the part of the ECA code which changes the GT crate mask with the idea that we should only ever have a crate out of the GT mask for a very good reason and that crate should probably stay out of the GT mask. @jcaravaca was there a reason you were resetting the GT crate mask?

### DIFF
--- a/Source/Experiments/SNOP/ECARun.h
+++ b/Source/Experiments/SNOP/ECARun.h
@@ -34,7 +34,6 @@
     int prev_coarsedelay;
     int prev_finedelay;
     uint16_t prev_pedwidth;
-    uint32_t prev_gtmask;
     uint32_t prev_pedmask;
 
     //ECA thread
--- a/Source/Experiments/SNOP/ECARun.m
+++ b/Source/Experiments/SNOP/ECARun.m
@@ -204,7 +204,7 @@ NSString* ORECARunFinishedNotification = @"ORECARunFinishedNotification";
         [anMTCModel setIsPedestalEnabledInCSR:TRUE];
 
         /* Enable Pedestal mask for all the crates */
-        [anMTCModel setPedCrateMask: 0x7FFFF];
+        [anMTCModel setPedCrateMask: (prev_pedmask | 0x7FFFF)];
     } @catch (NSException *e) {
         NSLogColor([NSColor redColor], @"setECASettings: error setting MTC model settings. name: %@ reason: %@.\n", [e name], [e reason]);
         goto err;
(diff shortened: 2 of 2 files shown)


jcaravaca:
I set both GTCrateMask and PedCrateMask to ensure we get pedestals in all crates, since it's very common to leave some crates out. There could be a case in which a crate is drop from the GTmask due to any reason or mistake, but we still want ECA constants.

Do you think the uninitialized object was the cause of ECA disabling TUB and CAEN? I'm trying to reproduce locally the issue with no success.

tlatorre-uchicago:
Yes, I suspect what happens is that since it's uninitialized, the prev_gtmask is set to 0, and then when it gets OR'ed with 0x7ffff it just sets it to 0x7ffff which leaves out TUBii.

I think it may be common to have the pedestal crate mask change (the nhit monitor changes it as well as ping crates), but the GT mask shouldn't ever change. If a crate is taken out of the GT mask the builder will essentially stop building events and that crate will be out of sync. The only case I can imagine us taking a crate out of the GT mask is if the XL3 is broken or there is some other problem which prevents us from getting good data from that crate.

jcaravaca:
Seems reasonable to me.

Also, you are not ORing the pedestal mask with the current pedestal mask. That'd leave the TUB and TUBii bits out of the mask in case they were enabled. So the question is whether those matter or not for ECAs...

tlatorre-uchicago:
I *think* they aren't physically connected, but not sure. If they were somehow connected we wouldn't want the pedestal to cause a GT on the CAEN because it would cause the CAEN to go out of sync. In general, in the nhit monitor for example, I've been just setting the pedestal mask directly and then reverting it after I'm done.

marzece:
The pedestal signal that gets sent to TUBii is not connected to anything. So it makes no difference whatsoever if tubii is in the ped crate mask. 

The same is not true totally true for the TUB. The TUB provides three things that use the pedestal signal, the first is a BNC on the front panel that is a copy of the ped signal, the second is the speaker (which has a knob so the speaker clicks on PEDs (or SYNC/SYNC24) instead of the more familiar GT). The third is a weird thing having to do with strobing LEDs that were on the PSUP during SNO.

To the best of my knowledge those three things are never used and are unlikely to ever be used. So all that is to say, it doesn't really matter if the TUB or TUBii are in the PED crate mask.

tlatorre-uchicago:
ok, hadn't realized that there was anything connected there. I've reverted that change.

tlatorre-uchicago:
merged.

PR #474, opened 2017-12-11
Update log Prefix on init

EdLeming:
Adds a method call to `[SNOPModel initWithCoder]` to prefix the log file name with the operator machine's hostname. 

This is for debugging purposes in cases where multiple ORCAs are running.


EdLeming:
I need to test on testand. Will drop a note here when that's done.

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -338,6 +339,22 @@ - (NSString *) xl3Host
     return xl3Host;
 }
 
+- (void) setLogPrefix
+{
+    NSArray *dataFileModels = [[(ORAppDelegate*)[NSApp delegate] document] collectObjectsOfClass:NSClassFromString(@"ORDataFileModel")];
```
EdLeming: @tlatorre-uchicago could you please provide some advice?

 This line always appears to return a null array. It's also true if I swap the ORDataFileModel string for ELLIEModel, ORMTCModel, ORXL3Model. 

It appears those models do not exist when initWithCoder is called on the SNOPModel. Would you expect that? There are a number of functions (e.g. setXL3Host) which get called earlier in initWithCoder than this new function I've written expecting these models to be available. I've checked the equivalent returned array in the call to setXL3Host (line 432) and it also gets set to NULL. 

This was true both locally and on the teststand.

tlatorre-uchicago: No, those objects aren't guaranteed to exist in initWithCoder. For the case of the XL3, MTC, etc. they update the hostnames and ports etc when they boot up, so this setting here is redundant.

If you want to make sure the object exists you should do it in AwakeAfterDocumentLoaded.

EdLeming:
This is now tested. The output can be seen on the teststand at /Users/snotdaq/2017/12/Logs/. The log for Run 15602 is named teststand.sp.snolab.ca_Run15702.log. 

How do people feel about that format? I had originally wanted to add the host as a suffix (rather than a prefix), but the code for doing so is significantly more complicated. There is a member variable which can be generically set for outfile prefix, but not the suffix (see below). Also, '.' in a file name?

https://github.com/EdLeming/orca/blob/log_hostname/Source/Objects/Data%20Processing/Data%20Storage/ORDataFileModel.h#L61

tlatorre-uchicago:
It would be a little bit more handy to have it as a suffix so that you could easily look at the different logs for a given run number by doing something like:

```bash
$ ls Run15702*
```

EdLeming:
OK, it now adds a suffix. I've had to make small changes to ORRunModel and ORDataFileModel to do so. The new output can again be seen on the testtand logs directoy:

`Run15707_teststand.sp.snolab.ca.log`

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -338,6 +339,30 @@ - (NSString *) xl3Host
...
+    if (![dataFileModels count]) {
+        NSLogColor([NSColor redColor], @"Must have a DataFileModel object in the configuration\n");
+        return;
+    }
+    ORDataFileModel* aDataFileModel = [dataFileModels objectAtIndex:0];
+    [aDataFileModel setFilePrefix:@""];
+    
+    NSArray *runObjects = [[(ORAppDelegate*)[NSApp delegate] document] collectObjectsOfClass:NSClassFromString(@"ORRunModel")];
+    if(![runObjects count]){
+        NSLogColor([NSColor redColor], @"detectorStateChanged: couldn't find run control object!");
```
tlatorre-uchicago: comment should be updated since it's not called when the detector state changes.

Source/Objects/Control/RunControl/ORRunModel.m
```diff
@@ -680,6 +682,19 @@ - (NSString*)dirName
     return dirName;
 }
 
+- (void) setFileSuffix:(NSString *)aFileSuffx
+{
+    [[[self undoManager] prepareWithInvocationTarget:self] setFileSuffix:[self fileSuffix]];
```
tlatorre-uchicago: The undo manager should only be called when a GUI action is performed. It looks like currently there is no way to set this from the GUI is that righit?

Source/Objects/Control/RunControl/ORRunModel.m
```diff
@@ -1761,6 +1776,7 @@ - (void) runStarted:(BOOL)doInit
 							  [NSNumber numberWithLong:runNumber],kRunNumber,
 							  [NSNumber numberWithLong:subRunNumber],kSubRunNumber,
 							  [NSNumber numberWithLong:[[ORGlobal sharedGlobal] runMode]],  kRunMode,
+                              [self fileSuffix], kFileSuffix,
```
tlatorre-uchicago: indentation?

Source/Objects/Data Processing/Data Storage/ORDataFileModel.m
```diff
@@ -949,7 +949,7 @@ - (NSString*) formRunName:(id)userInfo
 		if([filePrefix rangeOfString:@"Run"].location != NSNotFound){
 			s = [NSString stringWithFormat:@"%@%@%d",filePrefix,fileSuffix,runNumber];
 		}
-		else s = [NSString stringWithFormat:@"%@%@Run%@%d",filePrefix,[filePrefix length]?@"_":@"",fileSuffix,runNumber];
+		else s = [NSString stringWithFormat:@"%@%@Run%d_%@",filePrefix,[filePrefix length]?@"_":@"",runNumber,fileSuffix];
```
tlatorre-uchicago: If I understand this line correctly it will always insert an underscore right? We should definitely not make any changes that affect the default way the logs are stored as this could impact other experiments.

Source/Objects/Control/RunControl/ORRunModel.m
EdLeming: That's right. I put it in there with the thought that we might add this to the GUI at some point, but I guess if we do wanna do that we can just add it back in then.

EdLeming:
Yeah, I'll ping Mark Howe about this. I'll let you know the outcome.

EdLeming:
I had an e-mail conversation with Mark. I've updated based on his recommendations. He's now happy (Tony was cc'd)

tlatorre-uchicago:
One last thing to check is to make sure that the order here is correct. I actually don't remember exactly what happens when we set the run number etc. when Orca starts up. The code which does that is here: https://github.com/snoplus/orca/blob/a5c5ab5bb2bf03dab427de9fd651e4b45653e53f/Source/Experiments/SNOP/SNOPModel.m#L1165 and is called when the detector state is loaded. It would be good to check how that affects setting this log file, i.e. what happens when:

1. Orca boots up
2. Currently no run exists (so where do log files go?)

Then, there is the ordering of these two operations:

3. Log suffix is set
4. Run is set in in RunControlModel

I'm not sure what order they happen in and how 4. affects the log file that is written to.

EdLeming:
Hey @tlatorre-uchicago. So looking into this I found that, when ORCA opens up in the middle of an active run the logs will currently get saved as Run0.log. That file gets built when the run stops. 

In the last commit includes a fix to correctly set the run number in this case. The hostname will now also be included.

On a slight aside. Do we use the ORCA datafiles for anything? This PR includes a fix to make sure their filenames are also properly formatted in the same 'opening ORCA in the middle of a run' edge case. The code to do that is at the below. If you don't care about the datafiles, and you don't want to clutter up the code, I can remove that block. 

https://github.com/EdLeming/orca/blob/4d89be9b351ce44c822622b1124b9fd40c97624c/Source/Experiments/SNOP/SNOPModel.m#L1197

EdLeming:
One extra thing to note. Log files only get built at a run stop. Closing ORCA in the middle of a run will mean that the console output from the run start to the point of ORCA closing will be lost. 

I did think about catching the ORAboutToQuit (or whatever) notification and forcing the log to be saved when ORCA is closed. At the time I convinced myself we didn't want to do that, but I can't remember my reasoning. 

Do you think we should also add that catch?

tlatorre-uchicago:
Yeah, I think we should do that since ~12% of runs during normal physics running will be handovers.

Source/Objects/Data Processing/Data Storage/ORDataFileModel.m
```diff
@@ -947,11 +962,11 @@ - (NSString*) formRunName:(id)userInfo
...
 	if(filePrefix!=nil){
 		if([filePrefix rangeOfString:@"Run"].location != NSNotFound){
-			s = [NSString stringWithFormat:@"%@%@%d",filePrefix,fileSuffix,runNumber];
+			s = [NSString stringWithFormat:@"%@%@%d%@",filePrefix,fileSuffix,runNumber,fileStaticSuffix];
 		}
-		else s = [NSString stringWithFormat:@"%@%@Run%@%d",filePrefix,[filePrefix length]?@"_":@"",fileSuffix,runNumber];
+		else s = [NSString stringWithFormat:@"%@%@Run%@%d%@",filePrefix,[filePrefix length]?@"_":@"",fileSuffix,runNumber,fileStaticSuffix];
 	}
-	else s = [NSString stringWithFormat:@"Run%@%d",fileSuffix,runNumber];
+	else s = [NSString stringWithFormat:@"Run%@%d%@",fileSuffix,runNumber,fileStaticSuffix];
```
tlatorre-uchicago: In these lines if fileStaticSuffix is `nil`, you will get a segfault. It looks like for fileSuffix, this is checked for at the top, so you should probably do this here. I'm not 100% sure, but I think when you first add objects to the experiment, `initWithCoder` is not actually called, and so it's possible that fileStaticSuffix is `nil`.

Source/Objects/Control/RunControl/ORRunModel.m
```diff
@@ -1577,10 +1577,16 @@ - (void) finishRunStop
...
                                                       userInfo: statusInfo];
     //SV - End of changes
+
+    NSMutableDictionary* robustRunInfo = [NSMutableDictionary dictionaryWithCapacity:50];
+    if(runInfo){
+        robustRunInfo = runInfo;
+    } else { // If ORCA opened in the middle of a run, runInfo won't exist so must remake.
+        robustRunInfo = [NSMutableDictionary dictionaryWithDictionary:[self runInfo]]; //
+    }
+    [nextObject preCloseOut:robustRunInfo];
```
tlatorre-uchicago: I think it would be much cleaner to do something like this here:

```objective-c
if (!runInfo) {
    /* If Orca opens in the middle of a run, the runInfo object won't exist, so we must remake it. */
    runInfo = [self runInfo];
}
```

since there already exists a method for creating a runInfo object. Alternatively, we could fix this upstream by having the SNO+ model call some method to create it when we start the run control going in the middle of a run.

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -338,6 +339,50 @@ - (NSString *) xl3Host
...
+
+- (void) saveLogFiles:(NSNotification*)aNote
+{
+    /*
+     This method gets called by a notification observer, which waits for ORCA to close.
+    */
+    NSArray*  runObjects = [[(ORAppDelegate*)[NSApp delegate] document] collectObjectsOfClass:NSClassFromString(@"ORRunModel")];
+    if(![runObjects count]){
+        NSLogColor([NSColor redColor], @"waitForRunNumber: couldn't find run control object!");
+        [[NSNotificationCenter defaultCenter] postNotificationName:ORReleaseRunStateChangeWait object: self];
```
tlatorre-uchicago: Where in the new code do you post a run state change wait that needs to be cleared?

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -338,6 +339,50 @@ - (NSString *) xl3Host
...
+    NSArray *dataFileModels = [[(ORAppDelegate*)[NSApp delegate] document] collectObjectsOfClass:NSClassFromString(@"ORDataFileModel")];
+    if (![dataFileModels count]) {
+        NSLogColor([NSColor redColor], @"SetLogNameFormat: can't find ORDataFileModel!\n");
+        return; // (should never happen)
+    }
+    ORDataFileModel* aDataFileModel = [dataFileModels objectAtIndex:0];
+    
+    // Since ORCA's closing I'm not sure, if I post this note, that it will consistently be caught. Instead I'm gonna just pass it straight through.
+    NSNotification* myNote = [NSNotification notificationWithName:ORFlushLogsNotification object:self userInfo:[runControl runInfo]];
+    [aDataFileModel closeOutLogFiles:myNote];
```
tlatorre-uchicago: I believe notifications are posted immediately and on the same thread, so you should be fine to just post the notification here.

tlatorre-uchicago:
XCode is also warning that your declaration of `saveLogFiles` is incorrect.

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -1171,6 +1224,21 @@ - (void) detectorStateChanged:(NSNotification*)aNote
...
+        NSLogColor([NSColor redColor], @"SetLogNameFormat: can't find ORDataFileModel!\n");
+        return; // (should never happen)
+    }
+    ORDataFileModel* aDataFileModel = [dataFileModels objectAtIndex:0];
+
+    // Check if the datafile name is defined. If ORCA has just started up, it won't be
+    if([[aDataFileModel fileName] isEqualToString:@""]){
+        NSLog(@"SNOPModel: There was no datafile name!\n");
+        [self setLogNameFormat]; // this should have already been called by awakeAfterDocumentLoaded, but do it here too for piece of mind
+        [aDataFileModel runTaskStarted:[runControl runInfo]];
```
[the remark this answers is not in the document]
EdLeming: It sets the log file (and data file) paths on ORCA startup. They are automatically initialized to @"". If ORCA has previously gone through any run state change these will have been updated.

These lines force those paths to be updated such that all file IO will be available on ORCA startup exactly as if changing between any other type of run state.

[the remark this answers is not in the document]
tlatorre-uchicago: ok, probably worth adding a comment to that affect here since it's not obvious

Issue #472, opened 2017-12-07
Orca should re-send keep alives to tubii if it disconnects then re-connects.


marzece:
I was attempting to debug things on tubii and the pulsers wouldn't work b/c orca had given up on sending keep alive packets after I'd disconnected then reconnected tubii. So orca should probably re-start the keep-alives whenever tubii connects. Either that or put a large difficult to ignore warning if keep-alives are no longer being sent.

PR #476, opened 2017-12-15
Add AMELLIE GUI

EdLeming:
This long overdue PR updates the ELLIE gui to add functionality for AMELLIE. It also provides an option to re-start the heatbeat / keep alive to TUBii, which would resolve issue #472.  

Last time I managed to speak with the AMELLIE group they provided some requests for how this GUI should allow users to select fibres. Those requests are addressed. I've never had final conformation from the group that they are happy, but I think it's probably just time to get this pulled through.


Source/Experiments/SNOP/SNOPController.m
```diff
@@ -347,6 +357,16 @@ - (void) registerNotificationObservers
...
                         object: nil];
     
+    [notifyCenter addObserver :self
+                     selector : @selector(startAmellieRunNotification:)
+                         name : ORAMELLIERunStart
+                        object: nil];
+    
+    [notifyCenter addObserver :self
+                     selector : @selector(stopAmellieRunAction:)
+                         name : ORAMELLIERunFinished
```
tlatorre-uchicago: I realize it's already done this way for the tellie stuff, but it would be good to have the name include Notification to be consistent with everything else in Orca.

Source/Experiments/SNOP/SNOPController.m
```diff
@@ -1439,6 +1464,42 @@ -(void) fetchTellieRunFilesFinish:(NSNotification *)aNote
...
+    // Call getAmellieRunFiles from the model. This queries the DB and sets the tellieRunFiles
+    // property. This function runs asyncronously so we have to wait for a notification to be
+    // posted back before we can fill and re-activate the dropdown list (see below).
+    [model getAmellieRunFiles];
+}
+
+-(void) fetchAmellieRunFilesFinish:(NSNotification *)aNote
+{
+    // When we get a noticication that the database read has finished, set local variables
+    NSMutableDictionary *runFileDict = [[NSMutableDictionary alloc] initWithDictionary:[model amellieRunFiles]];
```
tlatorre-uchicago: Can use `[NSMutableDictionary dictionaryWithDictionary:[model amellieRunFiles]]` so that you don't have to release it.

Source/Experiments/SNOP/SNOPController.m
```diff
@@ -1828,6 +1886,198 @@ - (IBAction) stopTellieRunAction:(id)sender
...
+    // Check if a run is already ongoing
+    // If so tell the user and ignore this
+    // button press
+    if([[theELLIEModel tellieThread] isExecuting]){
+        NSLogColor([NSColor redColor], @"[AMELLIE]: A fire sequence is already on going. Cannot launch a new one until current sequence has finished\n");
+        return;
+    }
+
+     if(![[model lastStandardRunType] isEqualToString:@"AMELLIE"]){
+     ORRunAlertPanel(@"The AMELLIE standard run is not loaded.",@"You must load a AMELLIE standard run and start a new run before starting a fire sequence",@"OK",nil,nil);
```
tlatorre-uchicago: indentation.

Source/Experiments/SNOP/SNOPController.m
```diff
@@ -1828,6 +1886,198 @@ - (IBAction) stopTellieRunAction:(id)sender
...
+    NSUInteger nHits = [[[self amellieRunFile] objectForKey:@"NHits"] integerValue];
+    NSUInteger no_pulses = [[[self amellieRunFile] objectForKey:@"trigger_per_node"] integerValue];
+    NSUInteger rate = [[[self amellieRunFile] objectForKey:@"trigger_rate"] integerValue];
+    NSUInteger delay = [[[self amellieRunFile] objectForKey:@"trigger_delay"] integerValue];
+    NSUInteger slaveValue = [[[self amellieRunFile] objectForKey:@"slave_mode"] integerValue];
+    BOOL inSlave = NO;
+    if(slaveValue == 1){
+        inSlave = YES;
+    }
+    NSMutableArray* settingsArray = [[NSMutableArray alloc] init];
```
tlatorre-uchicago: Needs to be released or called with the autorelease version.

Source/Objects/Custom Hardware/SNO+/ELLIE/ELLIEController.m
```diff
@@ -423,6 +522,28 @@ - (IBAction)tellieBuildPushToDBAction:(id)sender {
...
+        return;
+    }
+        
+    [amellieFireButton setEnabled:NO];
+    [amellieStopButton setEnabled:YES];
+    [[NSNotificationCenter defaultCenter] postNotificationName:ORAMELLIERunStart object:nil userInfo:[self guiFireSettings]];
+    [amellieValidationStatusTf setStringValue:@""];
+}
+
+- (IBAction)amellieStopButton:(id)sender {
```
tlatorre-uchicago: Empty function?

Source/Objects/Custom Hardware/SNO+/ELLIE/ELLIEController.m
```diff
@@ -734,6 +855,120 @@ -(void) tellieBuildValidateAction:(id)sender
...
+    /*
+    if(![model amellieNodeMapping]){
+        [model loadAMELLIEStaticsFromDB];
+    }
+    //If still can't get reference, return
+    if(![model amellieNodeMapping]){
+        NSLogColor([NSColor redColor], @"[TELLIE]: Cannot connect to couchdb database\n");
+        return;
+    }
+    */
```
tlatorre-uchicago: Commented out code.

Source/Objects/Custom Hardware/SNO+/ELLIE/ELLIEController.m
```diff
@@ -1233,11 +1523,11 @@ -(NSString*)validateTellieNoPulses:(NSString *)currentText
 /////////////////////////////////////////////
 - (IBAction)telliePing:(id)sender {
     if([model pingTellie]){
-        NSString* response = [NSString stringWithFormat:@"Connected to tellie at:\n\n%@:%@", [model tellieHost], [model telliePort]];
+        NSString* response = [NSString stringWithFormat:@"Connected to tellie at: \%@:%@", [model tellieHost], [model telliePort]];
```
tlatorre-uchicago: Left in a backslash?

Source/Objects/Custom Hardware/SNO+/ELLIE/ELLIEModel.m
```diff
@@ -380,15 +394,26 @@ -(NSMutableDictionary*) returnTellieFireCommands:(NSString*)fibre withNPhotons:(
...
+    } else {
+        firePars = [self tellieFireParameters];
+    }
+    
     // Check if fire parameters have been successfully loaded
-    if([self tellieFireParameters] == nil){
-        NSLogColor([NSColor redColor], @"[TELLIE]: TELLIE_FIRE_PARMETERS doc has not been loaded from telliedb - you need to call loadTellieStaticsFromDB");
+    if(firePars == nil){
+        if(amellie){
+            NSLogColor([NSColor redColor], @"[AMELLIE]: TELLIE_FIRE_PARMETERS doc has not been loaded from telliedb - you need to call loadTellieStaticsFromDB");
```
tlatorre-uchicago: missing newline

Source/Objects/Custom Hardware/SNO+/ELLIE/ELLIEModel.m
```diff
@@ -380,15 +394,26 @@ -(NSMutableDictionary*) returnTellieFireCommands:(NSString*)fibre withNPhotons:(
...
+        } else {
+            NSLogColor([NSColor redColor], @"[TELLIE]: TELLIE_FIRE_PARMETERS doc has not been loaded from telliedb - you need to call loadTellieStaticsFromDB");
```
tlatorre-uchicago: same here.

Source/Objects/Custom Hardware/SNO+/ELLIE/ELLIEModel.m
```diff
@@ -1377,6 +1509,132 @@ -(void)parseTellieRunPlans:(id)aResult
...
+    [runDocDict setObject:subRunArray forKey:@"sub_run_info"];
+    
+    [self setAmellieRunDoc:runDocDict];
+    [[self couchDBRef:self withDB:@"amellie"] addDocument:runDocDict tag:kAmellieRunDocumentAdded];
+}
+
+- (void) updateAmellieRunDocument:(NSDictionary*)subRunDoc
+{
+    /*
+     Update [self tellieRunDoc] with subrun information.
```
tlatorre-uchicago: tellie or amellie?

Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiModel.m
```diff
@@ -950,11 +961,32 @@ -(void)pulseKeepAlive:(id)passed
...
+    // release memory
     [pool release];
 }
 
--(void)killKeepAlive
+-(void)restartKeepAlive:(NSNotification*)aNote{
+    /*
+     If the keep alive has died, as a user to re-start it.
+     */
+    dispatch_sync(dispatch_get_main_queue(), ^{
```
tlatorre-uchicago: Is this guaranteed to always be run from a separate thread? If so, might be good to document it in the comment above, otherwise Orca will hang.

tlatorre-uchicago: Also, how is this going to handle the case when TUBii is down? Is it going to keep running and popping up?

Source/Objects/Custom Hardware/SNO+/ELLIE/ELLIEController.m
```diff
@@ -777,25 +1012,38 @@ -(void)controlTextDidEndEditing:(NSNotification *)note {
...
+    
     //If still can't get reference, return
     if(![model tellieNodeMapping]){
         NSLogColor([NSColor redColor], @"[TELLIE]: Cannot connect to couchdb database\n");
         return;
     }
+
+    //If still can't get reference, return
+    if(![model amellieFibreMapping]){
+        NSLogColor([NSColor redColor], @"[AMELLIE]: Cannot connect to couchdb database\n");
```
tlatorre-uchicago: Is this the correct error message?

Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiModel.m
EdLeming: I've added a message to make that clearer. If TUBii's down the user can simply his cancel. If the user wants to restart at a later time they can do so via the servers tab in the ELLIE gui.

EdLeming:
Re: the last commit. I had a quick e-mail exchange with Mark Howe last week, resulting in him updating the ORCouchDB model to handle https (although it's not yet in our repo). It checks for a useHttps flag in any delegate class of an ORCouchDB object.

This commit adds that member variable to the ELLIEModel and includes checkbox in the ELLIE gui servers tab to allow toggling between http and https.

tlatorre-uchicago:
We discussed on the recent DAQ call that the current plan for ORCA is to just connect to the normal HTTP port when talking to CouchDB so we shouldn't need any changes for HTTPS right now. Could you separate out the HTTPS stuff into a separate pull request (it can be on top of the GUI changes here)?

tlatorre-uchicago:
Since both you and Javi have made changes to the SNO+ GUI you will have to coordinate on who is going to re-add the GUI elements on top of the other. Please let me know in which order you would like to merge and I will merge that first and then the other person can reset the SNOP.xib and re-add their stuff. @jcaravaca

tlatorre-uchicago:
XCode is giving the following warnings:

- Border Color property requires NSCustomBox type and NSLine Border type (a bunch of these in SNOP.xib)
- Instance method '-killKeepAlive' not found
- Potential leak of an object stored into `smellieFlaggingCli` (this one may not be due to this pull request)

tlatorre-uchicago:
Using XCode 9, I'm also getting the following warnings:

- The '_amellieRunFiles' ivar in 'SNOPModel' was retained by a synthesized property but not released before '[super dealloc]'
- The '_amellieRunNameLabel' ivar in 'SNOPModel' was retained by a synthesized property but not released before '[super dealloc]'
- (similar messages for SNOPController and various amellie variables)
- (similar messages for ELLIEModel) although these may not be new

EdLeming:
I don't have xcode 9, but running the analyser resulted in the same warnings. I've now resolved all instances I can find. 

Do you want me to squash this down or do you use the squash option when you merge?

tlatorre-uchicago:
I have to squash when I merge, so this is fine.

tlatorre-uchicago:
Still not sure 100% on what versions of XCode we are required to support (haven't gotten a definitive answer). But this current pull request fails to compile on XCode 6.2 which is what we've been using on the teststand. So, you might have to update the code to compile on this XCode, or if you can get in touch with Mark and get a definitive statement that we are only required to compile ORCA on XCode >= 7 then I can go ahead and update the XCode on the teststand.

EdLeming:
@tlatorre-uchicago, I've updated the .xibs to be compatible with 6.2. I'll quadruple check this is fine on the teststand before the leave the meeting.

tlatorre-uchicago:
There are a few small issues left:

1. Merge issue

It looks like the recent PIN readout pull request conflicts with this. I think you should just be able to merge with master, resolve the conflict, and push the merged branch.

2. SNOP.xib and ELLIE.xib

The SNOP.xib and ELLIE.xib files seem to have acquired a ton of <animations /> tags and previous color and font tags have been deleted. As far as I can tell, these changes are unrelated to the AMELLIE stuff. I'm also a little nervous because the XIB now has a lot of warnings when compiled on XCode 6.2 on the teststand.

Is the current version of the xib files created in XCode 6.2 starting from a fresh copy of the xib file on master?

Also, there are a ton of empty lines with whitespace. You can see them by enabling color and running:

```bash
$ git checkout master
$ git diff ...amellie_gui
```

EdLeming:
Yeah sorry about this, it's a real ball ache. It appears all those warnings are popping up as I used a more modern version of xcode when editing, which has added a load of `<animation/>` tags and systematically removed the below lines from the definition of NSBox objects.

```
<color key="borderColor" white="0.0" alpha="0.41999999999999998" colorSpace="calibratedWhite"/>
<color key="fillColor" white="0.0" alpha="0.0" colorSpace="calibratedWhite"/>
```

Apparently those methods are [not callable if the box is set to any type other than 'custom'](https://developer.apple.com/documentation/appkit/nsbox/1429797-fillcolor) . Throughout both GUI's they're set to [Primary](https://developer.apple.com/documentation/appkit/nsbox.boxtype). That being the case, logic would suggest it was the right thing to remove them... So I'm not certain when the compiler's complaining. Google doesn't have much to say about this issue.

I tried going through the diff and adding these lines back in, but xcode wouldn't build after. After some further unsuccessful playing around this afternoon I can't really think of a better solution than rebuilding both .xib files from master. That will take some annoying amount of time, I'll do it tomorrow.

EdLeming:
OK @tlatorre-uchicago, I think this is ready to go

tlatorre-uchicago:
Looks good. Just a few more spurious tabs in SNOPController, SNOPModel, ELLIEController.m, and ELLIEModel.m.

I see you added a few more commits dealing with couchdb error checking. Is this pull request something you can test locally on the detector before we deploy it, or do you think it's already good to go?

EdLeming:
I've been checking for whitespace additions with git diff master.. path/to/file, I can't see any new additions in those files. The diff on here also looks clean. Perhaps I'm only set to to look for tabs and you're thinking about spaces?

I tested the couchdb error checking by connecting to the VPN and pointing to the telliedb at dbug.sp.snolab.ca:6984. It worked as expected. I just tested on the teststand - the error checking works, but the couchdb queries are unsucessful. Putting that same address:port into a browser window on the teststand gives a 'site can't be reached' message, but is successful from my laptop when I'm on the VPN.  Is the teststand on a separate network?

tlatorre-uchicago:
Hm. When I do the following:

```bash
$ git checkout master
$ git diff ...amellie_gui
```

I get the following:

screen shot 2018-03-20 at 11 44 29 am

The teststand isn't on a separate network, but I disabled connections from the teststand to most of the underground computer to prevent people from accidentally controlling the detector when thinking they were just running on the teststand. So, currently you can't connect to many of the underground computers from the teststand.

EdLeming:
OK, I was reading the red as whitespace I'd removed compared to master - my mistake. I've got rid of all instances I could find. 

If the teststand can't talk to dbug then I'm happy everything makes sense.

tlatorre-uchicago:
merged.

PR #478, opened 2017-12-18
don't require EXT2 to be set high when running the nhit monitor and ping crates

tlatorre-uchicago:
Since we can't currently set EXT2 high this commit removes the code which sets EXT2 high before running the nhit monitor and ping crates. The solution for cutting potential events caused by sending pedestals is to cut events within 1 second of pedestal events which will need to be updated in RAT.

### DIFF
--- a/Source/Experiments/SNOP/NHitMonitor.m
+++ b/Source/Experiments/SNOP/NHitMonitor.m
@@ -416,8 +416,6 @@ err:
     int i;
     ORFec32Model *fec;
     int slot, channel;
-    uint32_t sync_trigger_mask, async_trigger_mask, tubii_dac;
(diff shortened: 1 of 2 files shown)


tannerbk:
Once TUBII is back to working are we going to want to go back to using the EXT2?

tlatorre-uchicago:
My thought was that since we already have a lot of runs without the EXT2 set high and we are going to have lots of runs again, it's better to just leave it completely out, i.e. since we are already going to have to deal with the case without it it's better not to complicate things and have it sometimes enabled and sometimes not, but I'll send out an email to the DAQ group to get other opinions.

tlatorre-uchicago:
merged.

PR #480, opened 2017-12-21
fix memory leak in SessionDB

mastbaum:
Release instance vars copied by synthesized properties inside `SessionDB dealloc` before `[super dealloc]`. I think this should correct the leak noticed by Mark (though I'm not 100% sure what the analysis was that found it, and Instruments is being super crashy).

### DIFF
--- a/Source/Experiments/SNOP/SessionDB.m
+++ b/Source/Experiments/SNOP/SessionDB.m
@@ -41,11 +41,11 @@ sessionKey = _sessionKey;
 {
     [connection disconnect];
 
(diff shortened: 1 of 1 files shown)


tlatorre-uchicago:
I think what Mark used to find those was XCode -> Analyze using XCode 8. If you have XCode 8, could you try running that and see if it comes up with anything?

mastbaum:
Updated based on Analyze in Xcode 8.

tlatorre-uchicago:
merged.

PR #481, opened 2018-01-05
SNO+: http->https

marzece:
I tested this on my laptop and it caused no issues. I bumped the timeout b/c I occasionally saw errors when testing this.

This also requires a change to the saved .orca file to change the port for orca/debugdb to 443


tlatorre-uchicago:
One issue with this change is that we won't be able to connect directly to a couchdb database which might be necessary if there are proxy issues or for example, on the teststand where we connect directly to a couchdb instance running on the local machine. One option would be to add a check box to use https, but perhaps it would be simpler to just have the http:// prefix be included in the hostname in the SNO+ control panel?

PR #482, opened 2018-01-10
Modal alert when missing run ID from DB

### LINKED ISSUE #168: Need more warning before starting run 0xffffffff
ORCA currently gets the new run number from the detector database when a run is started, but if the database operation fails a run will be started with number 0xffffffff, which will result in (effectively) a loss of data.  ORCA should probably post an app-modal dialog before starting the run if it was operator-initiated, but what to do on auto-started runs?

mastbaum:
Pop up a modal warning when unable to fetch the next run number from the database. This is in addition to the existing warning printed to the log.

screenshot 2018-01-10 15 17 44

Closes #168

### DIFF
--- a/Source/Experiments/SNOP/SNOPModel.m
+++ b/Source/Experiments/SNOP/SNOPModel.m
@@ -881,6 +881,14 @@ err:
 
     [runControl setRunNumber:0xffffffff - 1];
 
+    /* Pop up warning indicating that the default run number is being used. */
+    NSAlert* alert = [[[NSAlert alloc] init] autorelease];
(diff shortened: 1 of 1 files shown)


tlatorre-uchicago:
One issue discussed in #168, is what to do if it's a run rollover. Will this block the starting of the new run or not? I'm imagining the case where the shifter is going to the bathroom or not paying attention, and the run rolls over automatically. Will this prevent the new run from being started?

mastbaum:
After discussion, will implement this using the Orca alarm system rather than as a modal.

Issue #475, opened 2017-12-13
Orca inconsistent with detector state


nbarros:
After the GUI freeze reported in #464, when a new run was restarted Orca successfully started the new run, but reported as if it was stopped. 

![screen shot 2017-12-13 at 3 59 42 am](https://user-images.githubusercontent.com/1487800/33931084-fa9e29c6-dfbc-11e7-9ca1-aff7a9b4bafc.png)

Issue #450, opened 2017-09-15
Orca showing wrong run type


mmdepatie:
For runs 105048, 105049, Orca is showing MAINTENANCE as the run type, yet we are in PHYSICS as confirmed by detector state and the run type word.

mmdepatie:
![](https://files.slack.com/files-pri/T0SCP6RDH-F73RN6GCB/pasted_image_at_2017_09_15_02_01_pm.png)

![](https://files.slack.com/files-pri/T0SCP6RDH-F7513FRDM/pasted_image_at_2017_09_15_02_02_pm.png)

PR #485, opened 2018-02-01
SNO+: Gray out Standard Run popup menus

### LINKED ISSUE #484: Changing run in run type drop down while run is starting results in inconsitent run info
I was attempting to change the standard run values by switching into a physics run then quickly changing to a diagnostic run.  While ORCA was changin runs and the stoplight graphic was at yellow,  I changed the run selection drop down to select diagnostic. When the run actually started orca claimed the run type was diagnostic. But the run type word in the database for the run was physics. I was also able to change detector settings because orca though it was in a diagnostic run.

jcaravaca:
Attempt to fix #484

### DIFF
--- a/Source/Experiments/SNOP/SNOPController.m
+++ b/Source/Experiments/SNOP/SNOPController.m
@@ -787,6 +787,8 @@ snopGreenColor;
         [lightBoardView setState:kGoLight];
         [runStatusField setStringValue:@"Running"];
         [resyncRunButton setEnabled:true];
+        [standardRunPopupMenu setEnabled:true];
+        [standardRunVersionPopupMenu setEnabled:true];
         [runNumberField setStringValue:[runControl fullRunNumberString]];
         [doggy_icon start_animation];
 	}
@@ -795,6 +797,8 @@ snopGreenColor;
         [lightBoardView setState:kStoppedLight];
         [runStatusField setStringValue:@"Stopped"];
         [resyncRunButton setEnabled:false];
+        [standardRunPopupMenu setEnabled:true];
+        [standardRunVersionPopupMenu setEnabled:true];
         [doggy_icon stop_animation];
 	}
 	else if([runControl runningState] == eRunStarting || [runControl runningState] == eRunStopping || [runControl runningState] == eRunBetweenSubRuns){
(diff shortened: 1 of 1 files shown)


tlatorre-uchicago:
The fundamental problem here (which is maybe also responsible for #450 and #475) is that you have the display bound to lastStandardRunType, but this is not guaranteed to be the same as the standard run which was actually loaded to the detector because you set lastStandardRunType in the `runStarted` method which is *after* the settings were loaded.

So, I think it should be decided what exactly the `lastStandardRunType` variable should represent. If you want it to represent the last standard run that was loaded then it should be set in `loadStandardRun` (but note that this is not necessarily the same as actually running the run since you can load in a different standard run). If you want it to be the last standard run settings that were loaded at the start of a run, then it should go in `startStandardRun`, and finally, if you want it to be the last standard run that was *successfully* started, then you'll need a third variable like `pendingStandardRunType` or something which is set in `startStandardRun` and then `lastStandardRunType` is set to this variable when the run successfully starts.

jcaravaca:
The definition is the third one: `lastStandardRunType` should be the last SR that successfully started. As I understand your proposal, it is equivalent than graying out the menu so that it's not possible to change the current `standardRunType`, right?. `pendingStandardRunType` would be equivalent to `standardRunType` in this case.

tlatorre-uchicago:
Sort of. This fix would probably work in 99% of the cases, but the standardRunType is set programatically (so it's not true that the only way to change it is through the GUI). For example, the standard run type variable is also set here: https://github.com/snoplus/orca/blob/6d59fa7575ff02f734d759db6d5c0d6654e2cef0/Source/Experiments/SNOP/SNOPModel.m#L2397. So, if for whatever reason some piece of code called `startStandardRun` in the middle of a run starting, it would display the wrong value. In addition, it's a bit of future proofing in case the standard run variable gets set from code somewhere else (I think there are various parts of the ELLIE code that automatically start runs without the user directly hitting the Start button).

Since this change is orthogonal to the issue above, though I'm happy to merge this since it will probably cover most of the cases. Let me know if you want to go ahead, and if so, I can merge after it's been tested.

jcaravaca:
Tested OK at the TS.

tlatorre-uchicago:
merged.

PR #486, opened 2018-02-05
Orca alarm when missing run ID from DB

mastbaum:
Post an Orca alarm when unable to fetch the next run number from the database. This is in addition to the existing warning printed to the log, but meant to be more attention-grabbing (but non-blocking).

Example of the alarm dialog window, with the detail drawer expanded:
screenshot 2018-02-05 10 42 10
An audible beep also sounds periodically until the alarm is acknowledged.

### DIFF
--- a/Source/Experiments/SNOP/SNOPModel.m
+++ b/Source/Experiments/SNOP/SNOPModel.m
@@ -505,6 +505,9 @@ tellieRunFiles = _tellieRunFiles;
     [_smellieRunFiles release];
     [_tellieRunFiles release];
     [_tellieRunNameLabel release];
+
+    [defaultRunAlarm clearAlarm];
+    [defaultRunAlarm dealloc];
     [super dealloc];
 }
 
@@ -632,6 +635,8 @@ tellieRunFiles = _tellieRunFiles;
     ORXL3Model *xl3;
     TUBiiModel *tubii;
     int i;
+
+    [defaultRunAlarm clearAlarm];
 
     objs = [[(ORAppDelegate*)[NSApp delegate] document]
          collectObjectsOfClass:NSClassFromString(@"ORMTCModel")];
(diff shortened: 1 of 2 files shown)


mastbaum:
Incidentally, I noticed I can't see the alarm dialog nib with Xcode 7.3.1:
screenshot 2018-02-05 10 24 19
Should we bump development (including the test stand Xcode) to version 8?

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -881,6 +881,14 @@ - (void) waitForRunNumber: (ORPQResult *) result
 
     [runControl setRunNumber:0xffffffff - 1];
 
+    /* Post an Orca alarm indicating that the default run number is being used. */
+    ORAlarm* defaultRunAlarm = [[ORAlarm alloc] initWithName:[NSString stringWithFormat:@"Using default run number"]
```
tlatorre-uchicago: Need to release this somewhere?

tlatorre-uchicago:
I think we can bump the version we use to build or develop on at will, it's not crucial (in fact, the most recent version is 9.2 on High Sierra), but we shouldn't update the teststand version. Although there aren't any clear minimum versions that we are supposed to support for Orca, I think Mark Howe still wants to keep support for OSX 10.6 and so it's good to make sure we can still build using older XCodes.

mastbaum:
Now I keep a pointer to the `ORAlarm` in `SNOPModel` -- pattern is the same as e.g. `ORRunModel`. It will also clear automatically if the run number is fetched successfully.

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -633,6 +636,8 @@ - (void) runInitialization:(NSNotification*)aNote
     TUBiiModel *tubii;
     int i;
 
+    [defaultRunAlarm clearAlarm];
```
tlatorre-uchicago: Does this need to check that it's initialized?

Source/Experiments/SNOP/SNOPModel.m
```diff
@@ -505,6 +505,9 @@ - (void) dealloc
     [_smellieRunFiles release];
     [_tellieRunFiles release];
     [_tellieRunNameLabel release];
+
+    [defaultRunAlarm clearAlarm];
+    [defaultRunAlarm dealloc];
```
tlatorre-uchicago: Same here, should check that it's actually allocated.

mastbaum:
I don't think such a check is necessary. [Apple Obj-C docs](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/WorkingwithObjects/WorkingwithObjects.html) say:
> ... it’s perfectly acceptable in Objective-C to send a message to nil. If you do send a message to nil, obviously nothing happens.

This is also follows the pattern of `ORRunModel`.

tlatorre-uchicago:
merged.

PR #489, opened 2018-02-22
Add new timeout handling to pollTellieFibre

EdLeming:
A couple of line change for handling the case where TELLIE receives <100% of the triggers requested in a flash sequence. 

This has been observed as a problem during some recent slave mode tests with TUBii. To properly handle this case the TELLIE hardware must be told to reset it's counters in order to appropriately handle the following sequence.

### DIFF
--- a/Source/Objects/Custom Hardware/SNO+/ELLIE/ELLIEModel.m
+++ b/Source/Objects/Custom Hardware/SNO+/ELLIE/ELLIEModel.m
@@ -997,7 +997,7 @@ err:
                 NSLogColor([NSColor redColor], errorString);
                 goto err;
             }
-    }
+        }
         
         ////////////////////
         // Init can take a while. Make sure no-one hit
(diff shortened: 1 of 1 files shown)


mnirkko:
Looks ok to me.

Source/Objects/Custom Hardware/SNO+/ELLIE/ELLIEModel.m
```diff
@@ -975,7 +997,7 @@ -(void) startTellieRun:(NSDictionary*)fireCommands
                 NSLogColor([NSColor redColor], errorString);
                 goto err;
             }
-        }
+    }
```
tlatorre-uchicago: Typo?

tlatorre-uchicago:
merged (note: won't be deployed yet, so keep an eye out for when that happens if this requires changes to be deployed on the tellie side at the same time).

mnirkko:
Awesome, thanks. It would be helpful to have this by Monday, so we (Matt & I) can test it while running TELLIE/TUBii underground.

tlatorre-uchicago:
I don't think a new release will be made by then, but you can always compile the master version of Orca and run it "locally".

PR #492, opened 2018-03-13
ELLIE: Add retain calls in setters to avoid bad releases in dealloc

EdLeming:
Mark Howe pointed out that some custom setters in the ELLIE model were causing ORCA to crash when closing the experiment. I have reproduced locally. This PR fixes the problem.  

The setters were overridden so they can be easily called from the Controller - which sets model variables when the GUI is updated. I found the model wasn't updating as expected when using the autogenerated setters.


tlatorre-uchicago:
merged.

PR #494, opened 2018-03-22
TELLIE: Add checkbox for tuning runs

EdLeming:
This feature allows an operator to flag 'tuning runs' while TELLIE channels are being tuned for PCA intensity settings. To date the group has observed these settings drift on the day-week scale, so tuning is a more regular occurrence than originally expected. 

When the flag is active the run won't roll over at the end of the requested flash sequence. This should help us stop burning through so many runs (which I think nearline and processing don't respond very well to).

### DIFF
--- a/Source/Objects/Custom Hardware/SNO+/ELLIE/ELLIEController.h
+++ b/Source/Objects/Custom Hardware/SNO+/ELLIE/ELLIEController.h
@@ -62,8 +62,9 @@
     IBOutlet NSPopUpButton *tellieExpertFibreSelectPb;
     IBOutlet NSPopUpButton *tellieExpertOperationModePb; //Operation mode (master or slave)
     
+    IBOutlet NSButtonCell *tellieExpertTuningCb;
     IBOutlet NSTextField *tellieExpertValidationStatusTf;
-    
+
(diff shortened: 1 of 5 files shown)


tlatorre-uchicago:
Looks good to me. The only thing you might want to adjust is the position. At least on the teststand, the "Tuning run" text is cut off by the edge of the well and so it looks like "Tuning ru".

One other thing unrelated that I noticed when testing on the teststand is that when I switch from Fire Fibre -> Build Custom Sequence as the "General operator" the tab doesn't get highlighted correctly, i.e. the Fire Fibre tab stays highlighted. Not sure why this would be, perhaps it's XCode getting confused because there are three nested layers of tabs.

EdLeming:
Thanks @tlatorre-uchicago, I've updated the text to avoid things cutting off. 

You're right about the tab highlighting - it only highlights correctly on the second click. I've found a few threads online about issues with nested tabs, but can't find anything on this explicitly. I tried changing the inheritance of the appearance characteristics, but that didn't fix it. Seeing as it doesn't affect functionality I'll put it up as an issue and have a look at it when I've got some more time.

tlatorre-uchicago:
merged.

PR #497, opened 2018-04-03
SMELLIE Crash fix

EdLeming:
Resolves issue #495. Crash was happening at line 2164. I couldn't initially reproduce in local testing as the run control wasn't running, so the fireSettingsArray wasn't being checked in my local version.

This PR moves the declaration of that variable to before the first goto to make absolutely sure the object is properly constructed - not just some random memory address on the stack. I put in the check at line 2163 in mind of catching this case, but hadn't considered that the compiler might assign the memory without putting the correct object type there. I guess this is why people don't like goto's :(


tlatorre-uchicago:
merged.

PR #500, opened 2018-04-19
Change allowed values for TELLIE triggerdelay to any positive integer

EdLeming:
I've had a request to update the allowed range of values for the TELLIE trigger delay. Values are currently allowed in the 0-1275 ns range in 5ns steps. This is the based on the specs of the delay chip on the TELLIE hardware. However, since the group expect to be running in slave mode (through TUBii) they have asked that I replace this check with a positive integer check. They have specifically asked that it have no defined top end.


EdLeming:
I didn't test on the detector, but I did on the teststand. This PR only changes auto values for the TELLIE gui so I don't anticipate any problems.

PR #506, opened 2018-05-16
CRSU debugging: empty strings 

MRiganSUSX:
Simple fix to insert an empty string whenever an empty variable is passed back. 
Not critical, but was throwing assertion failures in Xcode when testing. 
This might be also present at other places, these were found during short test on teststand.

### DIFF
--- a/Source/Objects/Custom Hardware/SNO+/Fec32/ORFec32Controller.m
+++ b/Source/Objects/Custom Hardware/SNO+/Fec32/ORFec32Controller.m
@@ -546,7 +546,11 @@
 
 - (void) commentsChanged:(NSNotification*)aNote
 {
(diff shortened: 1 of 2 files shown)


MRiganSUSX:
Hi @tlatorre-uchicago :
These errors were seen in the Xcode console whenever a 'FEC card view' window or a 'Crate view' window was closed. No output to Orca log itself.
(sorry don't have orca control right now to do screenshots, but these are just the very standard windows whenever you click a crate and specific fec inside crate).

tlatorre-uchicago:
Ok, I am able to reproduce this. I don't understand exactly why these methods (for example, boardIDChanged) are being called when the board id has a null value. This seems like an issue which should be fixed "upstream" of these methods so the notifications aren't posted when the value is NULL.

PR #508, opened 2018-05-30
Fix TUBii orphans

EdLeming:
This PR removes the problem runAboutToStart function from the TUBii model and moves the registration of notification observers to initWithCoder. 

A single line is also added to the keepAliveFunction so the ELLIE gui properly updates when the keepAlive is no longer running. The time between pulses is also increase to 5s in order to avoid 'flooding' the TUBii server with more request then are required. I checked with @icoulter and the server itself expects a ping every 30s, so this is still well within tolerance. 

@tlatorre-uchicago, if it's possible it might be a good idea to test this on the detector later when you're deploying the new servers.

### DIFF
--- a/Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiModel.m
+++ b/Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiModel.m
@@ -104,6 +104,8 @@ NSString* ORTubiiSettingsChangedNotification    = @"ORTubiiSettingsChangedNotifi
 #pragma mark •••Archival
 - (id) initWithCoder:(NSCoder *)aCoder {
     self = [super initWithCoder:aCoder];
+    [self registerNotificationObservers];
+    [self activateKeepAlive];
     if (self) {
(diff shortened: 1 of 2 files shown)


Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiModel.m
```diff
@@ -104,6 +104,8 @@ - (void) dealloc {
 #pragma mark •••Archival
 - (id) initWithCoder:(NSCoder *)aCoder {
     self = [super initWithCoder:aCoder];
+    [self registerNotificationObservers];
+    [self activateKeepAlive];
```
tlatorre-uchicago: I think you probably don't want to activate this until after you've loaded the settings. The connection to the tubii server for example, isn't set up until later in this function.

EdLeming:
Good point. It worked on the teststand, but there's gotta be a race going on there. I've moved to post client allocation.

boardhead:
Merged

PR #512, opened 2018-06-15
Only push one TELLIE couchdb document per run

EdLeming:
The TELLIE 'Tuning Run' tick box was added in #494. This PR updates the flash sequence logic such that only one couchdb document is generated for a run when this box is checked. 

That single document will contain the hardware settings sent to the TELLIE hardware (and PIN diode readback) for each sub-run contained within a 'Tuning Run'.


EdLeming:
We still need to test #508 on the detector. Seeing as that's the case I'd like like to test this on the detector at the same.

boardhead:
I have merged this into the Orca64 branch.

PR #514, opened 2018-07-14
Prevent tubii run start exception from bubbling up

### LINKED ISSUE #505: cannot start runs without TUBii
ORCA should be able to start runs without TUBii, but on the teststand currently it seems as if you cannot start runs without TUBii. To reproduce, open the TUBii GUI and change the hostname to a non existent hostname and try to start a run. You should get the following error:

```
Run Not Started because of exception: RedisClient, reason: Operation timed out
```

I'm *pretty* sure that this used to work. @EdLeming any idea if any recent changes might have affected this?

marzece:
When starting a run tubii's runAboutToStart function gets called and if it raises an uncaught exception the run will be prevented from starting. I think we don't in general want this to be the case so this commit just catches and warns about any exceptions that occur instead of letting them stop the run from starting. Fixes #505

### DIFF
--- a/Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiModel.m
+++ b/Source/Objects/Custom Hardware/SNO+/TUBii/TUBiiModel.m
@@ -178,9 +178,16 @@ NSString* ORTubiiSettingsChangedNotification    = @"ORTubiiSettingsChangedNotifi
                        object : nil];
 }
 - (void) runAboutToStart: (NSNotification*) aNone {
-    [self setDataReadout:NO];
-    [self ResetFifo]; //Maybe take this out eventually? I'm not sure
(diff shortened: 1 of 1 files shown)


boardhead:
I think this PR is not applicable once PR#508 has been applied.

ecaden:
Hi Phil,

When will it be applied?

Thanks,
Erica 

--
Erica Caden, Ph.D.
SNOLAB Research Scientist
SNO+ Detector Manager
ecaden@snolab.ca
+1.705.692.7000 x2253

tannerbk:
@boardhead we cannot run the detector currently, so this or #508 should go on in ASAP.

boardhead:
Should not be necessary since #508 is now applied

PR #516, opened 2018-08-29
	64-bit version of Orca

boardhead:
Initial testing on teststand and detector at low voltage has been successful.  Next step is to try handing off to 32-bit Orca, and if successful, running at HV with operators underground for a few shifts.  Then a careful review of the data before using for production datataking.


boardhead:
Merged.

PR #526, opened 2019-02-04
two small fixes

tannerbk:
- new lines to nhit monitor messages
- grey out the ping crates button during a physics run

That ping crates button is actually kind of confusing, since it doesn't run the "standard" ping crates. I think it could still be potentially useful still though.


boardhead:
combined into 1.1.2 update 34d00d04d36dc82660e3e0fa8ac2f14beb24e818

PR #529, opened 2019-06-07
	Added additional validation of run number loaded from the database

boardhead:
Requires testing on test stand


boardhead:
Tested on test stand, and it works great.

PR #532, opened 2021-07-08
Smellie mode update

jlidgard:
For an upcoming SMELLIE test, we’d like to use SMELLIE in ‘slave’ mode – (i.e. where we accept a trigger rather than generating our own). SMELLIE was designed with the option to run in this mode, but we haven’t for a very long time because we hadn’t needed to (and not all our hardware is compatible with this mode). However, the functionality to select this mode has evolved out of ORCA. This PR aims to bring this mode back in.

There’s an ‘if’ statement in ELLIEModel.m (line 2081) which checks to see if the laser field is ‘superK’ or ‘else’ – this is the current limitation. In this PR I add a further if statement to select either ‘master’ or ‘slave’ mode.
The extra statement checks against a new field read from a SMELLIE couchDB document: 'trigger_mode'

The setSmellieLaserHeadSlaveMode function requires a time argument, this is based on the existing couchDB fields: 'nTriggers'/'rate'*1.2+60 (here adding an extra 1minute for a possible laser switch time, and +20% for leeway). So when using slave mode these fields should be filled with the expected no of triggers and rate.

Note at this stage I have not tested to see if this compiles. I have to find a mac to do that. I'm submitting the code for pre-review (so do not merge yet)
@dcookman @tsengj10 @reichold @BenLand100

### DIFF
--- a/Source/Objects/Custom Hardware/SNO+/ELLIE/ELLIEModel.m
+++ b/Source/Objects/Custom Hardware/SNO+/ELLIE/ELLIEModel.m
@@ -2124,6 +2124,7 @@ err:{
 
                         @try{
                             NSNumber *withEstimatedTime = @([nTriggers integerValue]/[rate floatValue]*1.2 + 60);
+							NSLog(@"[SMELLIE]: Timeout slave mode time\t\t\t: %1.1f seconds\n", [withEstimatedTime floatValue]);
                             [self setSmellieLaserHeadSlaveMode:laserSwitchChannel withIntensity:intensity withFibreInput:fibreInputSwitchChannel withFibreOutput:fibreOutputSwitchChannel withTime:withEstimatedTime withGainVoltage:gain];
                         } @catch(NSException* e){
                             NSLogColor([NSColor redColor], @"[SMELLIE]: Problem with smellie server request: %@\n", [e reason]);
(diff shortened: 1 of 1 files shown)


jlidgard:
@boardhead @tannerbk  I'm trying to compile Orca but getting errors. I'm running Xcode 11.3.1 - is that ok?
I've cloned this repo, I checked out the 'drivers' from SVN (i did this after reading the readme), selected 'Product -> Run'... and I get a 'Error: multiple commands produce '(long directory path)...' error. 
Any advice on getting this to build to test my PR?

BenLand100:
Hi @jlidgard I don't believe the steps in the readme are relevant for SNO+ orca, since all of our hardware is controlled over the network (no special drivers). I don't think this would cause the issue you're having, but maybe worth building the repo as-is to check.

jlidgard:
I tried building without the drivers folder and same issue.. I'm worried I need a specific xcode version, is that true? (Or was that for the old 32bit version?)
________________________________

jlidgard:
There's also many 'build' options within xcode (build, test, run, run plans..) It could be I need the right option selected, I'll try a few..
________________________________

EdLeming:
Hi All - looks like I'm still subscribed to this repo. I'm gonna change
that now, but seeing as I've seen this probably the best thing to do is
test on the surface lab test stand. Then you know problems in compilation
aren't associated with your local install.

reichold:
Hi Ed
Would you mind exchanging Email addresses so we can stay in touch for non business stuff like parties and pints and the like?
Cheers
Armin

jlidgard:
I've now tested this compiles on the test stand (I found and fixed one small bug). I believe the next step would be to run this branch during a shift and test it with SMELLIE.

jlidgard:
Archive this. Don't pull until needed (if ever).

